PR #97, opened 2015-02-10
Log complete command line into logger

lzap:
Several times I struggled with finding out what parameter user/customer used.
This should add it to the log file. Untested.


ares:
Few comments, 

1. args may not be useful since default values are modified by current answer file (first vs second run)
1. should be probably in debug log level
1. some nicer wrapping message would be fine, like triggered with these args
1. I think we should try to log args sooner
1. if you're interested in what was specified on CLI, why not using `ARGV` instead of args?

ares:
ping @lzap

lzap:
Hmmm DEBUG level is useless for my case. Closing.

PR #98, opened 2015-03-02
Fixes #9509 - allow overriding by empty values

ares:


domcleal:
Not the prettiest thing, but I guess it's good!  The only improvement that comes to mind is supporting a --no-class-parameter argument, like the way we disable entire modules.

ares:
then we'd have the same issue in interactive mode, and it's consistent with UNDEF we already have there

domcleal:
I think you can set an empty array in interactive mode already though.

ares:
when I tested I was unable to do that, it always got back to default value

domcleal:
Ah alright.

mmoll:
:+1:

ares:
Thanks both of you, merging.

PR #99, opened 2015-03-03
limit highline version to ruby 1.8 compatible one

mmoll:


ares:
Merging, thanks @mmoll!

PR #102, opened 2015-03-08
Fixes #7881 - add logs when default values fetching fails

ares:


ares:
[test]

domcleal:
Untested, but looks sane.

ares:
Thanks, merging

PR #106, opened 2015-03-09
Fixes 6279 - add skip system checks option

ares:


README.md
```diff
@@ -711,7 +711,8 @@ but you want the logs to be readable by specific users.
 When you want to make sure that user has some software installed or has the
 right version you can write a simple script and put it into checks directory.
 All files found there will be ran and if any has non-zero exit code, kafo
-wont execute puppet.
```
mmoll: I'm not native English speaker, but i think a slightly better wording would be:

All files found there will be executed and if any of these exits with an non-zero exit code, kafo won't execute puppet but print an error message `Your system does not meet configuration criteria`.

README.md
```diff
@@ -723,6 +724,11 @@ java -version 2>&1 | grep OpenJDK
 exit $?
 ```
 
+If you want to ignore results of check scripts, you can use builtin
```
mmoll: "you can use _the_ builtin"

ares:
Thanks @mmoll, would you give it another round please?

mmoll:
untested, but visually OK :+1:

ares:
Thank you, merging.

PR #108, opened 2015-04-01
Fixes #9996: Exposes Puppet's --profile as an option for installers.

ehelms:


ares:
Would you mind adding docs to https://github.com/theforeman/kafo/blob/master/README.md probably under Advanced topics and explain what it's good for? Also I can't find --profile docs anywhere and it seems it's not present on 2.7.25, could you also cover supported versions in that README please?

ehelms:
Added a section to the README with a link to the post I originally found (https://puppetlabs.com/blog/tune-puppet-performance-profiler) on the topic.

ares:
Thank you @ehelms, merging!

PR #113, opened 2015-05-26
Enable progress bar with Puppet 3.8

mmoll:
Maybe this check should get removed anyway... ;)

### DIFF
--- a/modules/kafo_configure/lib/puppet/parser/functions/add_progress.rb
+++ b/modules/kafo_configure/lib/puppet/parser/functions/add_progress.rb
@@ -2,7 +2,7 @@ require 'kafo_configure/lib/kafo/puppet/report_wrapper'
 
 module Puppet::Parser::Functions
   newfunction(:add_progress) do |args|
-    supported = %w(2.6. 2.7. 3.0. 3.1. 3.2. 3.3. 3.4. 3.5. 3.6. 3.7.)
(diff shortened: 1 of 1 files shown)


ares:
@mmoll I guess we're just lucky that the monkey patch still works :-)
rerunning [test], I added puppet 3.8 to matrix

modules/kafo_configure/lib/puppet/parser/functions/add_progress.rb
```diff
@@ -2,7 +2,7 @@
 
 module Puppet::Parser::Functions
   newfunction(:add_progress) do |args|
-    supported = %w(2.6. 2.7. 3.0. 3.1. 3.2. 3.3. 3.4. 3.5. 3.6. 3.7.)
+    supported = %w(2.6. 2.7. 3.0. 3.1. 3.2. 3.3. 3.4. 3.5. 3.6. 3.7. 3.8)
```
ares: actually `3.8` would also match 3.80, would you mind adding dot at the end? (3.80 is never going to happen, but because of [OCD](http://en.wikipedia.org/wiki/Obsessive%E2%80%93compulsive_disorder)...)

mmoll: copypasto :cry:

ares:
Merging, thanks @mmoll!

PR #115, opened 2015-06-09
Add pre_commit hook between param wizard/validation and answer file writing

domcleal:


ares:
I think you need to list it here as well https://github.com/domcleal/kafo/blob/pre-commit-hook/lib/kafo/hooking.rb#L11 so they are loaded automatically.

domcleal:
Good catch, thanks.  Tested as well now, seems OK.

ares:
Thanks @domcleal, I appreciate improving doc for `pre` hook as well. Merging!

PR #116, opened 2015-06-10
Treat any return value from stdlib's validate_* functions as success

domcleal:
validate_*, in particular, validate_re and validate_integer do not
return values, they are in fact statements or assertions.

Any error thrown from these functions should be treated as false, and
any successful execution of the function is a success.


ares:
Thanks @domcleal, merging!

PR #117, opened 2015-08-19
Fixes #10162 - Added scenario support

mbacovsky:
The PR contains implementation of scenarios. Things that are missing follows:
 * [x] tests
 * [x] code clean up
 * [ ] ~~handling of installation with different scenario (exit vs. answer file merge)~~ will be addressed in separate PR
 * [x] support for existing config file foremat (so that existing foreman_installer can run with new Kafo)
 * [x] KafoWizards packaging https://github.com/theforeman/kafo_wizards

Summary of changes:
 * Foreman installer binary should be replaced with 
```ruby
#!/usr/bin/env ruby
require 'rubygems'
require 'kafo'

# where to find scenarios
CONFIG_DIR = '/etc/foreman/installer-scenarios.d/'

# Run the install
@result = Kafo::KafoConfigure.run

# handle exit code when help was invoked or installer ended with '2' (success in puppet)
if @result.nil? || (!@result.config.app[:detailed_exitcodes] && @result.exit_code == 2)
  exit(0)
else
  exit(@result.exit_code)
end
```
this is dropping support for config files in ./config. Do we want it?

 * config directory was changed to 
```
/etc/foreman
├── installer-scenarios.d
│   ├── foreman-installer-answers.yaml
│   ├── foreman-installer.yaml
│   ├── foreman-proxy-answers.yaml
│   ├── foreman-proxy.yaml
│   └── last_scenario.yaml -> /etc/foreman/installer-scenarios.d/foreman-proxy.yaml
...
```
 * Config file format changes:
   * `:modules_dir` was renamed to `:module_dirs` and should be array. (`:modules_dir` still works)
   * `:name` is mandatory name of the scenario
   * `:description` optional scenario description
   * `:check_dirs` was added accepting list of dirs with checks, defaults to current (installer_root)/checks 

Mapping stays unchanged, installer is looking for the manifests in the module_dirs (first occurence is used). I hope puppet does the same later :) (needs to be checked)  

 * There are two new `foreman-installer` options and the behavior changed slightly
```
$> foreman-installer
Scenario was not selected, can not continue. Use --list-scenarios to list available options.
```

List of scenarios
```
$> foreman-installer --list-scenarios
Available scenarios
  Foreman (use: --scenario foreman-installer)
        Basic and most generic installation of Foreman
  Foreman Proxy (use: --scenario foreman-proxy)
        Install Foreman proxy without Foreman
```

Installer help
```
$> foreman-installer --scenario foreman-proxy -h
Usage:
    foreman-installer [OPTIONS]

Options:

= Generic:
     ...
    --list-scenarios LIST_SCENARIOS List available installation scenaraios
    ...
```

Installer in interactive mode
```
$> foreman-installer -i

Select installation scenario

Please select one of the pre-set installation scenarios. You can customize your installtion later during the installtion.

Available actions:
1. Foreman: Basic and most generic installation of Foreman
2. Foreman Proxy: Install Foreman proxy without Foreman
3. Cancel Installation
Your choice: 
```

Sample installer run
```
$> foreman-installer --scenario foreman-proxy
 /Stage[main]/Foreman_proxy::Register/Foreman_smartproxy[scenarios.my.lan]: Could not evaluate: Proxy scenarios.my.lan cannot be registered (401 Unauthorized): N/A
Installing             Done                                               [100%] [...............................................................................................................................]
  * Foreman Proxy is running at https://scenarios.my.lan:8443
  The full log is at /var/log/foreman-installer/foreman-installer.log
```

### DIFF
--- a/lib/kafo/configuration.rb
+++ b/lib/kafo/configuration.rb
@@ -13,6 +13,8 @@ module Kafo
     end
 
     DEFAULT = {
+        :name                 => '',
+        :description          => '',
         :log_dir              => '/var/log/kafo',
         :log_name             => 'configuration.log',
         :log_level            => 'info',
@@ -20,7 +22,7 @@ module Kafo
         :mapping              => {},
         :answer_file          => './config/answers.yaml',
         :installer_dir        => '.',
-        :modules_dir          => './modules',
+        :module_dirs          => ['./modules'],
         :default_values_dir   => '/tmp',
         :colors               => Configuration.colors_possible?,
         :color_of_background  => :dark,
@@ -104,7 +106,7 @@ module Kafo
         temp_dir = Dir.mktmpdir(nil, app[:default_values_dir])
         KafoConfigure.exit_handler.register_cleanup_path temp_dir
         @logger.info 'Loading default values from puppet modules...'
-        command = PuppetCommand.new("$temp_dir=\"#{temp_dir}\" #{includes} dump_values(#{params})", ['--noop']).append('2>&1').command
+        command = PuppetCommand.new("$temp_dir=\"#{temp_dir}\" #{includes} dump_values(#{params_to_dump})", ['--noop']).append('2>&1').command
         result = `#{command}`
         @logger.debug result
         unless $?.exitstatus == 0
@@ -145,6 +147,23 @@ module Kafo
       File.open(filename, 'w') { |file| file.write(config_header + format(YAML.dump(data))) }
     end
 
+    def params
+      @params ||= modules.map(&:params).flatten
+    end
+
+    def preset_parameters
+      # set values based on default_values
+      params.each do |param|
+        param.set_default(params_default_values)
+      end
+
+      # set values based on YAML
+      params.each do |param|
+        param.set_value_by_config(self)
+      end
+      params
+    end
+
     private
 
     def custom_storage
@@ -153,16 +172,18 @@ module Kafo
 
     def includes
       modules.map do |mod|
-        params_file = File.join(KafoConfigure.modules_dir, mod.params_path)
-        @logger.debug "checking presence of #{params_file}"
-        File.exist?(params_file) ? "include #{mod.dir_name}::#{mod.params_class_name}" : nil
+        module_dir = KafoConfigure.module_dirs.find do |dir|
+          params_file = File.join(dir, mod.params_path)
+          @logger.debug "checking presence of #{params_file}"
+          File.exist?(params_file)
+        end
+        module_dir ? "include #{mod.dir_name}::#{mod.params_class_name}" : nil
       end.uniq.compact.join(' ')
     end
 
-    def params
-      params = modules.map(&:params).flatten
-      params = params.select { |p| p.default != 'UNSET' }
-      params.map { |param| "#{param.dump_default}" }.join(',')
+    def params_to_dump
+      parameters = params.select { |p| p.default != 'UNSET' }
+      parameters.map { |param| "#{param.dump_default}" }.join(',')
     end
 
     def format(data)
--- a/lib/kafo/exit_handler.rb
+++ b/lib/kafo/exit_handler.rb
@@ -15,7 +15,9 @@ module Kafo
           :manifest_error => 22,
           :no_answer_file => 23,
           :unknown_module => 24,
-          :defaults_error => 25
+          :defaults_error => 25,
+          :unknown_scenario => 26,
+          :scenario_error => 27
       }
     end
 
--- a/lib/kafo/kafo_configure.rb
+++ b/lib/kafo/kafo_configure.rb
@@ -21,6 +21,7 @@ require 'kafo/puppet_command'
 require 'kafo/progress_bar'
 require 'kafo/hooking'
 require 'kafo/exit_handler'
+require 'kafo/scenario_manager'
 
 module Kafo
   class KafoConfigure < Clamp::Command
@@ -28,8 +29,8 @@ module Kafo
 
     class << self
       attr_accessor :config, :root_dir, :config_file, :gem_root, :temp_config_file,
-                    :modules_dir, :kafo_modules_dir, :verbose, :app_options, :logger,
-                    :exit_handler
+                    :module_dirs, :kafo_modules_dir, :verbose, :app_options, :logger,
+                    :check_dirs, :exit_handler, :scenario_manager
       attr_writer :hooking
 
       def hooking
@@ -40,11 +41,17 @@ module Kafo
     def initialize(*args)
       self.class.logger           = Logger.new
       self.class.exit_handler     = ExitHandler.new
+      self.class.scenario_manager = ScenarioManager.new((defined?(CONFIG_DIR) && CONFIG_DIR) || (defined?(CONFIG_FILE) && CONFIG_FILE))
+
+      # Handle --list-scenarios before we need them
+      self.class.scenario_manager.list_available_scenarios if ARGV.include?('--list-scenarios')
+
       self.class.config_file      = config_file
       self.class.config           = Configuration.new(self.class.config_file)
       self.class.root_dir         = File.expand_path(self.class.config.app[:installer_dir])
-      modules_dir                 = self.class.config.app[:modules_dir] || (self.class.config.app[:installer_dir] + '/modules')
-      self.class.modules_dir      = File.expand_path(modules_dir)
+      self.class.check_dirs       = [self.class.config.app[:check_dirs] || File.join(self.class.root_dir, 'checks')].flatten
+      module_dirs                 = self.class.config.app[:modules_dir] || self.class.config.app[:module_dirs] || (self.class.config.app[:installer_dir] + '/modules')
+      self.class.module_dirs      = [module_dirs].flatten.map { |dir| File.expand_path(dir) }
       self.class.gem_root         = File.join(File.dirname(__FILE__), '../../')
       self.class.kafo_modules_dir = self.class.config.app[:kafo_modules_dir] || (self.class.gem_root + '/modules')
       @progress_bar               = nil
@@ -108,6 +115,7 @@ module Kafo
         store_params(temp_config_file)
       else
         store_params
+        self.class.scenario_manager.link_last_scenario(self.class.config_file) if self.class.scenario_manager.configured?
       end
       run_installation
       return self
@@ -216,6 +224,8 @@ module Kafo
       self.class.app_option ['-v', '--verbose'], :flag, 'Display log on STDOUT instead of progressbar'
       self.class.app_option ['-l', '--verbose-log-level'], 'LEVEL', 'Log level for verbose mode output',
                             :default => 'info'
+      self.class.app_option ['-S', '--scenario'], 'SCENARIO', 'Use installation scenario'
+      self.class.app_option ['--list-scenarios'], :flag, 'List available installation scenaraios'
     end
 
     def set_options
@@ -375,6 +385,7 @@ module Kafo
 
     def config_file
       return CONFIG_FILE if defined?(CONFIG_FILE) && File.exists?(CONFIG_FILE)
+      return self.class.scenario_manager.select_scenario if self.class.scenario_manager.configured?
       return '/etc/kafo/kafo.yaml' if File.exists?('/etc/kafo/kafo.yaml')
       return "#{::RbConfig::CONFIG['sysconfdir']}/kafo/kafo.yaml" if File.exists?("#{::RbConfig::CONFIG['sysconfdir']}/kafo/kafo.yaml")
       File.join(Dir.pwd, 'config', 'kafo.yaml')
--- a/lib/kafo/scenario_manager.rb
+++ b/lib/kafo/scenario_manager.rb
@@ -1,0 +1,137 @@
+# encoding: UTF-8
+require 'kafo_wizards'
+
+module Kafo
+  class ScenarioManager
+    attr_reader :config_dir, :last_scenario_link, :previous_scenario
+
+    def initialize(config, last_scenario_link_name='last_scenario.yaml')
+      @config_dir = File.file?(config) ? File.dirname(config) : config
+      @last_scenario_link = File.join(config_dir, last_scenario_link_name)
+      @previous_scenario = File.realpath(last_scenario_link) if File.exists?(last_scenario_link)
+    end
+
+    def available_scenarios
+      # assume that *.yaml file in config_dir that has key :name is scenario definition
+      @available_scenarios ||= Dir.glob(File.join(config_dir, '*.yaml')).reject { |f| f =~ /#{last_scenario_link}$/ }.inject({}) do |scns, scn_file|
+        begin
+          content = YAML.load_file(scn_file)
+          if content.is_a?(Hash) && content.has_key?(:answer_file)
+            # add scenario name for legacy configs
+            content[:name] = File.basename(scn_file, '.yaml') unless content.has_key?(:name)
+            scns[scn_file] = content
+          end
+        rescue Psych::SyntaxError => e
+          warn "Warning: #{e}"
+        end
+        scns
+      end
+    end
+
+    def list_available_scenarios
+      say "Available scenarios"
+      available_scenarios.each do |config_file, content|
+        scenario = File.basename(config_file, '.yaml')
+        say "  #{content[:name]} (use: --scenario #{scenario})"
+        say "        " + content[:description] unless (content[:description].nil? || content[:description].empty?)
+      end
+      say "  No available scenarios found in #{config_dir}" if available_scenarios.empty?
+      KafoConfigure.exit(0)
+    end
+
+
+    def scenario_selection_wizard
+      wizard = KafoWizards.wizard(:cli, 'Select installation scenario',
+        :description => "Please select one of the pre-set installation scenarios. You can customize your installtion later during the installtion.")
+      f = wizard.factory
+      available_scenarios.keys.each do |scn|
+        label = available_scenarios[scn][:name].to_s
+        label += ": #{available_scenarios[scn][:description]}" if available_scenarios[scn][:description]
+        wizard.entries << f.button(scn, :label => label, :default => true)
+      end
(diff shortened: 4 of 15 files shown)


ares:
> this is dropping support for config files in ./config. Do we want it?

ideally not, a long term goal is to be able to generate stand alone installers which can't put their answer files in `/etc`, otoh if it helps it could be explicit configuration option or env variable

lib/kafo/configuration.rb
```diff
@@ -20,7 +22,7 @@ def self.colors_possible?
         :mapping              => {},
         :answer_file          => './config/answers.yaml',
         :installer_dir        => '.',
-        :modules_dir          => './modules',
+        :module_dirs          => ['./modules'],
```
ares: how do we fix existing foreman installations? I wonder if `module_dirs` should default to `[$modules_dir]` somehow or use it if it's not defined

lib/kafo/configuration.rb
```diff
@@ -153,9 +155,12 @@ def custom_storage
 
     def includes
       modules.map do |mod|
-        params_file = File.join(KafoConfigure.modules_dir, mod.params_path)
-        @logger.debug "checking presence of #{params_file}"
-        File.exist?(params_file) ? "include #{mod.dir_name}::#{mod.params_class_name}" : nil
+        module_dir =  KafoConfigure.module_dirs.find do |dir|
```
ares: nit: double space after =

lib/kafo/exit_handler.rb
```diff
@@ -10,12 +10,13 @@ def initialize
 
     def error_codes
       @error_codes ||= {
-          :invalid_system => 20,
-          :invalid_values => 21,
-          :manifest_error => 22,
-          :no_answer_file => 23,
-          :unknown_module => 24,
-          :defaults_error => 25
+          :invalid_system   => 20,
```
ares: luckily this is not puppet, don't align hash rockets, btw @iNecas will be mad for breaking this :-) make sure you document this in [README](https://github.com/theforeman/kafo#exit-code)

lib/kafo/kafo_configure.rb
```diff
@@ -375,11 +399,83 @@ def unset
...
+          end
+        end
+      end
+      exit(0)
+    end
+
+    def prev_config_file
+      begin
+        File.realpath(last_scenario_link)
+      rescue Exception => e
```
ares: `rescue StandardError` or just `rescue => e` so we don't rescue from SyntaxError etc

lib/kafo/kafo_configure.rb
```diff
@@ -38,13 +39,20 @@ def hooking
     end
 
     def initialize(*args)
+
```
ares: no empty lines here please

lib/kafo/kafo_configure.rb
```diff
@@ -216,6 +225,9 @@ def set_app_options
       self.class.app_option ['-v', '--verbose'], :flag, 'Display log on STDOUT instead of progressbar'
       self.class.app_option ['-l', '--verbose-log-level'], 'LEVEL', 'Log level for verbose mode output',
                             :default => 'info'
+      self.class.app_option ['-S', '--scenario'], 'SCENARIO', 'Display log on STDOUT instead of progressbar'
+      self.class.app_option ['--list-scenarios'], 'LIST_SCENARIOS', 'Display log on STDOUT instead of progressbar'
+
```
ares: no empty lines here please

lib/kafo/kafo_configure.rb
```diff
@@ -375,11 +399,83 @@ def unset
...
+        Dir.glob(File.join(CONFIG_DIR, '*.yaml')).each do |config_file|
+          content = YAML.load_file(config_file)
+          if content.has_key?(:name)
+            scenario = File.basename(config_file, '.yaml')
+            say "  #{content[:name]} (use: --scenario #{scenario})"
+            say "        " + content[:description] if content[:description]
+          end
+        end
+      end
+      exit(0)
```
ares: use `self.class.exit(0)` here

lib/kafo/kafo_configure.rb
```diff
@@ -216,6 +225,9 @@ def set_app_options
       self.class.app_option ['-v', '--verbose'], :flag, 'Display log on STDOUT instead of progressbar'
       self.class.app_option ['-l', '--verbose-log-level'], 'LEVEL', 'Log level for verbose mode output',
                             :default => 'info'
+      self.class.app_option ['-S', '--scenario'], 'SCENARIO', 'Display log on STDOUT instead of progressbar'
```
ares: description does not seem to be correct

lib/kafo/kafo_configure.rb
```diff
@@ -216,6 +225,9 @@ def set_app_options
...
+      self.class.app_option ['--list-scenarios'], 'LIST_SCENARIOS', 'Display log on STDOUT instead of progressbar'
```
ares: description does not seem to be correct

lib/kafo/kafo_configure.rb
```diff
@@ -293,6 +305,18 @@ def store_params(file = nil)
...
+    def store_last_scenario
+      link_path = last_scenario_link
+      if link_path
+        File.delete(link_path) if File.exist?(link_path)
+        File.symlink(self.class.config_file, link_path)
+      end
+    end
+
+    def last_scenario_link
+      File.join(CONFIG_DIR, 'last_scenario.yaml') if defined?(CONFIG_DIR) && Dir.exists?(CONFIG_DIR)
```
ares: `'last_scenario.yaml'` should be constant I think

lib/kafo/kafo_configure.rb
```diff
@@ -375,11 +399,83 @@ def unset
...
       File.join(Dir.pwd, 'config', 'kafo.yaml')
     end
 
+    def select_scenario(config_dir)
+      # try scenario provided in the args via -S or --scenario
+      parsed = ARGV.join(" ").match /(--scenario|-S)(\s+|[=]?)(\S+)/
+      if parsed
+        scenario_file = File.join(config_dir, "#{parsed[3]}.yaml")
+        return scenario_file if File.exists?(scenario_file)
+        puts "Scenario (#{scenario_file}) was not found, can not continue"
```
ares: you could use logger with fatal level here I think

lib/kafo/kafo_configure.rb
```diff
@@ -375,11 +399,83 @@ def unset
...
+        self.class.exit(:unknown_scenario)
+      end
+
+      # try last_scenario.yaml
+      return self.class.prev_config_file if self.class.prev_config_file
+
+      # assume that in CONF_DIR *.yaml that has key :name is scenario definition
+      scenarios = Dir.glob(File.join(CONFIG_DIR, '*.yaml')).inject({}) do |scns, scn_file|
+        content = YAML.load_file(scn_file)
```
ares: load_file can raise exceptions if some file contains syntax error, so it should be rescued

lib/kafo/kafo_configure.rb
```diff
@@ -375,11 +399,83 @@ def unset
...
+        res =  wizard.run
+        if res == :cancel
+          say 'Installation was cancelled by user'
+          self.class.exit(1)
+        end
+        return res
+      end
+
+      # fail
+      puts "Scenario was not selected, can not continue. Use --list-scenarios to list available options."
```
ares: logger?

lib/kafo/kafo_configure.rb
```diff
@@ -375,11 +399,83 @@ def unset
...
+      self.class.exit(:unknown_scenario)
+    end
+
+    def self.list_available_scenarios
+      say "Available scenarios"
+      if defined?(CONFIG_DIR) && Dir.exists?(CONFIG_DIR)
+        Dir.glob(File.join(CONFIG_DIR, '*.yaml')).each do |config_file|
```
ares: this should be extracted to some method as the same thing is being done in `select_scenario` method

ares:
I think in general it's a good start. I see why you have to touch ARGV and can't reuse later parsing. I'd prefer if you could break the long methods into smaller isolated ones or even to a separate class. I didn't test it yet but looks promising.

lib/kafo/exit_handler.rb
```diff
@@ -10,12 +10,13 @@ def initialize
...
+          :invalid_values   => 21,
+          :manifest_error   => 22,
+          :no_answer_file   => 23,
+          :unknown_module   => 24,
+          :defaults_error   => 25,
+          :unknown_scenario => 26
```
iNecas: what about `wrong_scenario` or `undef_scenario`? :D

mbacovsky:
Update with some comments fixed. Existing config file format is supported, existing tests were fixed. Some comments need more investigation, tests for scenarios needs to be added.

mbacovsky:
`kafo_wizards` packaging PR: https://github.com/theforeman/foreman-packaging/pull/807

lib/kafo/configuration.rb
mbacovsky: Fixed, new Kafo should work with old config format

mbacovsky:
PR was updated, All comments were addressed. I moved scenarios stuff to ScenarioManager class to avoid polution of KafoConfigure. Some basic tests were added.

ares:
Could you please extend readme file, describe new features, config options and ideally a section describing how to add new scenario?

lib/kafo/kafo_configure.rb
```diff
@@ -194,6 +202,9 @@ def set_parameters
       params.each do |param|
         param.set_value_by_config(config)
       end
+
+      # set values based on last used scenario if it has been different
+      # TODO
```
ares: should this be implemented in this PR?

lib/kafo/kafo_configure.rb
```diff
@@ -108,6 +115,7 @@ def execute
         store_params(temp_config_file)
       else
         store_params
+        self.class.scenario_manager.link_last_scenario(self.class.config_file) if defined?(CONFIG_DIR)
```
ares: could you extract `defined?(CONFIG_DIR)` to some scenario manager method? it's used on multiple places.

ares:
needs rebase

lib/kafo/scenario_manager.rb
```diff
@@ -0,0 +1,123 @@
...
+          :description => "You are trying to replace existing installation with different scenario. This may lead to unpredictable states. Please confirm that you want to proceed.")
+        wizard.entries << wizard.factory.button(:proceed, :label => 'Proceed with selected installation scenario', :default => false)
+        wizard.entries << wizard.factory.button(:cancel, :label => 'Cancel Installation', :default => true)
+        result = wizard.run
+        if result == :cancel
+          say 'Installation was cancelled by user'
+          KafoConfigure.exit(0)
+        end
+      else
+        if (ARGV & ['--force']).empty?
```
ares: `if ARGV.include?('--force')`

lib/kafo/kafo_configure.rb
```diff
@@ -40,11 +41,17 @@ def hooking
     def initialize(*args)
       self.class.logger           = Logger.new
       self.class.exit_handler     = ExitHandler.new
+      self.class.scenario_manager = ScenarioManager.new(CONFIG_DIR || CONFIG_FILE)
```
ares: `CONFIG_DIR` might not be defined at this time (old bin script), if it's defined but set to nil/false it would crash on undefined `CONFIG_FILE` on new bin script I think

lib/kafo/kafo_configure.rb
```diff
@@ -375,6 +385,7 @@ def unset
 
     def config_file
       return CONFIG_FILE if defined?(CONFIG_FILE) && File.exists?(CONFIG_FILE)
+      return self.class.scenario_manager.select_scenario if self.class.scenario_manager.configured?
```
ares: this can cause fatal error if there's no scenario, unfortunately it applies even in `--help` and `-h` runs, since this is called early at https://github.com/theforeman/kafo/pull/117/files#diff-657b82e4fb9c80f11badef608b14de10L43

lib/kafo/scenario_manager.rb
```diff
@@ -0,0 +1,123 @@
...
+      @last_scenario_link = File.join(config_dir, last_scenario_link_name)
+      @previous_scenario = File.realpath(last_scenario_link) if File.exists?(last_scenario_link)
+    end
+
+    def available_scenarios
+      # assume that *.yaml file in config_dir that has key :name is scenario definition
+      @available_scenarios ||= Dir.glob(File.join(config_dir, '*.yaml')).reject { |f| f =~ /#{last_scenario_link}$/ }.inject({}) do |scns, scn_file|
+        begin
+          content = YAML.load_file(scn_file)
+          scns[scn_file] = content if content.has_key?(:name)
```
ares: The if here makes name value in config file mandatory, on old answer file this might not be present. At least we should make https://github.com/theforeman/kafo/blob/master/config/kafo.yaml.example#L7 uncommented and document it. Kafofy luckily generates the name for quite long time (since 0.4) so such old configurations might not exist widely.

lib/kafo/kafo_configure.rb
```diff
@@ -216,6 +224,8 @@ def set_app_options
       self.class.app_option ['-v', '--verbose'], :flag, 'Display log on STDOUT instead of progressbar'
       self.class.app_option ['-l', '--verbose-log-level'], 'LEVEL', 'Log level for verbose mode output',
                             :default => 'info'
+      self.class.app_option ['-S', '--scenario'], 'SCENARIO', 'Use installation scenario'
+      self.class.app_option ['--list-scenarios'], 'LIST_SCENARIOS', 'List available installation scenaraios'
```
ares: this does not accept parameter so no `'LIST_SCENARIOS'` I think

lib/kafo/scenario_manager.rb
```diff
@@ -0,0 +1,123 @@
...
+
+    def scenario_changed?(scenario)
+      !!previous_scenario && scenario != previous_scenario
+    end
+
+    def configured?
+      defined?(CONFIG_DIR)
+    end
+
+    def select_scenario
```
ares: this should log selected scenario for easier debugging

lib/kafo/scenario_manager.rb
```diff
@@ -0,0 +1,123 @@
...
+      scenario = scenario_from_args || previous_scenario ||
+        (available_scenarios.keys.count == 1 && available_scenarios.keys.first) ||
+        select_scenario_interactively
+      if scenario.nil?
+        KafoConfigure.logger.fatal  "Scenario was not selected, can not continue. Use --list-scenarios to list available options."
+        KafoConfigure.exit(:unknown_scenario)
+      else
+        # check if scenario was changed
+        confirm_scenario_change(scenario) if scenario_changed?(scenario)
```
ares: quite edge case but I hit it... if you run installer from directory in which you use some symlink it gets dereferenced for previous scenario because of https://github.com/theforeman/kafo/pull/117/files#diff-a7539345f28ead836346cbcefeb9debbR11 but the current scenario does not resulting in false change detection

I suppose we should wrap scenarion in `File.realpath($...)` as well

ares:
`--no-colors` seems to be ignored for early logging (e.g. Scenario (...) was not found, can not continue) or for interactive mode to select scenario of last was not found

lib/kafo/scenario_manager.rb
```diff
@@ -0,0 +1,123 @@
...
+      KafoConfigure.exit(0)
+    end
+
+
+    def scenario_selection_wizard
+      wizard = KafoWizards.wizard(:cli, 'Select installation scenario',
+        :description => "Please select one of the pre-set installation scenarios. You can customize your installtion later during the installtion.")
+      f = wizard.factory
+      available_scenarios.keys.each do |scn|
+        wizard.entries << f.button(scn, :label => "#{available_scenarios[scn][:name]}: #{available_scenarios[scn][:description]}", :default => false)
```
ares: prints name with colon even if description is blank, should be just name in such case

ares:
nit: when selecting scenario interactively, items are red, is this expected or hard to change? it's not consistent with other color scheme meaning, I'd expect grey or green color

lib/kafo/scenario_manager.rb
mbacovsky: Fixed in `scenario_changed?`

mbacovsky:
- rebased
- `--no-collors` in early logging fixed. 
- colors in nteractive mode fixed
- inline comments fixed except for `:name` docs and ability to print help with no scenario selected

lib/kafo/kafo_configure.rb
mbacovsky: I changed the way scenario manager prints the errors. It should be more user friendly now. It prints the message using `say` and logs the same in `error` debug level for cases when logging is set up already. 

I found implementation of printing `--help` with no config file non-trivial. If it is acceptable I would fix it in PR that will follow and will include refactoring bringing in better isolation of `KafoConfigure` and `Configuration` classes. My plan is to pass along Configuration objects rather then the config_file filenames, which would allow us to create e.g.  "dummy" configuration with no modules to finish `KafoConfigure` initialization and printing help. It should make testing easier as well.

lib/kafo/scenario_manager.rb
mbacovsky: I changed it to detect scenario files by `:answer_file` entry that I hope is mandatory and more distinctive. I'm guessing the scenario name from filename if it is missing. Is there any better candidate for scenario distinction?

mbacovsky:
- fixed issue with mandatory :name 
- fixed error printing
- commented on printing help with no config

lib/kafo/scenario_manager.rb
ares: I'm not aware of any

mbacovsky:
Kafofy and README were updated

ares:
LGTM, could you squash it please?

mbacovsky:
README updated, commits squashed.
@ares thanks a 1e6 for a quick response :beers:

ares:
Merging, thanks @mbacovsky for a huge patch. This will be the reason for 0.7.0!

PR #112, opened 2015-05-14
Fixed 1 failed test in wizard_test.rb

johnkim76:
Fixes the 1 failure when running tests, which yields the following results:

Finished tests in 0.961535s, 220.4807 tests/s, 400.4012 assertions/s.

  1) Failure:
Kafo::Wizard::menu navigation::#main_menu::list all modules, run installation and cancel options, then cancel#test_0001_displays menu [/vagrant/kafo/test/kafo/wizard_test.rb:85]:
Expected "\nMain Config Menu\n1. [y] Configure puppet\n2. Display current config\n3. Save and run\n4. Cancel run without Saving\nChoose an option from the menu... You must choose one of [1, 2, 3, 4].\n?  " to include "[✓] Configure puppet".

212 tests, 385 assertions, 1 failures, 0 errors, 0 skips
Coverage report generated for Unit Tests to /vagrant/kafo/coverage. 834 / 1202 LOC (69.38%) covered.
rake aborted!
Command failed with status (1): [ruby -I"lib:lib:test" -I"/home/vagrant/.rvm/gems/ruby-1.9.3-p551/gems/rake-10.4.2/lib" "/home/vagrant/.rvm/gems/ruby-1.9.3-p551/gems/rake-10.4.2/lib/rake/rake_test_loader.rb" "test/kafo/color_scheme_test.rb" "test/kafo/condition_test.rb" "test/kafo/exit_handler_test.rb" "test/kafo/help_builders/advanced_test.rb" "test/kafo/help_builders/basic_test.rb" "test/kafo/hook_context_test.rb" "test/kafo/hooking_test.rb" "test/kafo/logger_test.rb" "test/kafo/param_test.rb" "test/kafo/params/array_test.rb" "test/kafo/params/hash_test.rb" "test/kafo/params/password_test.rb" "test/kafo/puppet_module_test.rb" "test/kafo/validator_test.rb" "test/kafo/wizard_test.rb" ]
/home/vagrant/.rvm/gems/ruby-1.9.3-p551/bin/ruby_executable_hooks:15:in `eval'
/home/vagrant/.rvm/gems/ruby-1.9.3-p551/bin/ruby_executable_hooks:15:in `<main>'
Tasks: TOP => test
(See full trace by running task with --trace)

PR #118, opened 2015-08-25
bump logging and highline upper version limits

mmoll:
As newer Debian/Ubuntu versions are already bringing newer versions in their upstream OS repos, this is required to not run into problems regarding packaging and is related to issue 11280 for hammer/apipie-bindings.

I tested this with logger 2.0.0 and highline 1.7.2 - logging and progress bars seem unaffected.

### DIFF
--- a/kafo.gemspec
+++ b/kafo.gemspec
@@ -27,11 +27,11 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'kafo_parsers'
   spec.add_dependency 'puppet', '< 4.0.0'
   # better logging
-  spec.add_dependency 'logging', '< 2.0.0'
+  spec.add_dependency 'logging', '< 3.0.0'
   # CLI interface
   spec.add_dependency 'clamp', '>= 0.6.2'
   # interactive mode
-  spec.add_dependency 'highline', '>= 1.6.21', '< 1.7'
+  spec.add_dependency 'highline', '>= 1.6.21', '< 2.0'
   # ruby progress bar
   spec.add_dependency 'powerbar'
 end
--- a/Gemfile
+++ b/Gemfile
@@ -2,3 +2,11 @@ source 'https://rubygems.org'
 
 # Specify your gem's dependencies in kafo.gemspec
 gemspec
+
+  if RUBY_VERSION >= '2.0'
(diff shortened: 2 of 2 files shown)


domcleal:
I think there are a couple of issues with tests: a) some encoding issue on 1.9, b) gem still needs pinning in the Gemfile for 1.8

mmoll:
It's a bit better now, I have no idea why Ruby 1.9.3 is the only failing...

mmoll:
BZ 1245716 seems like a similar problem

kafo.gemspec
```diff
@@ -26,12 +26,19 @@ Gem::Specification.new do |spec|
   # puppet module parsing
   spec.add_dependency 'kafo_parsers'
   spec.add_dependency 'puppet', '< 4.0.0'
-  # better logging
-  spec.add_dependency 'logging', '< 2.0.0'
   # CLI interface
   spec.add_dependency 'clamp', '>= 0.6.2'
-  # interactive mode
-  spec.add_dependency 'highline', '>= 1.6.21', '< 1.7'
+  if RUBY_VERSION >= '1.9'
```
domcleal: You'll need to add any pinnings for 1.8 in a Gemfile (so it only affects development/test) rather than the gemspec, since when it's built as a binary gem then these conditionals won't stay in the final gem.

domcleal:
It might be a US-ASCII string coming out of another library or something, e.g. Puppet or rdoc.  It might take a bit of rooting around with a debugger..

mmoll:
It seems about the ✓ sign and might be related to #112, unfortunately I have no idea how to fix this, as this suggests that the detection of UTF-8 support is going wrong somewhere. @ares?

mmoll:
Ihere were quite some PRs in highline 1.7.0 that "fix" encoding things, so something of that will probably cause this...

mmoll:
Indeed, adding back the line 716 that got removed in davispuh/highline@6fba5f7 seems to fix the tests

mmoll:
As the only OSes that are relevant here are on 2.0 or 2.1, I'm going for Ruby >=2.0 now.
Build is :green_heart:

ares:
Thanks @mmoll, merging!

PR #119, opened 2015-08-28
Fixes output test on Ruby 1.9

ares:
fixes the issue described in #112


mmoll:
I tested this while debugging #118 and can confirm it works. :+1:

PR #120, opened 2015-08-28
Spec change for 0.6.12

ares:


rubygem-kafo.spec
```diff
@@ -13,13 +13,15 @@ URL: https://github.com/theforeman/kafo
 Source0: http://rubygems.org/downloads/%{gem_name}-%{version}.gem
 %if "%{?scl}" == "ruby193" || (0%{?rhel} == 6 && 0%{!?scl:1})
 Requires: %{?scl_prefix}ruby(abi)
+Requires: %{?scl_prefix}rubygem(logging) < 2.0.0
```
domcleal: I think you'd simply want the same version numbers in both parts of the conditional, as this part isn't used exclusively say for Ruby 1.8.  The 1.8 compatibility ought to be handled by whoever packages rubygem(logging).

ares:
thanks @domcleal for looking, so this should be enough? I suppose noone will bump highline in 1.9 SCL (which would break kafo while spec file allows that)

domcleal:
Because it'd break the non-SCL version with Ruby 1.8?  Yeah, that's fine I think.

ares:
> Because it'd break the non-SCL version with Ruby 1.8?

No, because in 1.9 SCL Ruby 1.9 would be used and Kafo does not work with highline 1.7 on Ruby 1.9. That's why I wanted to pin highline only for 1.9 SCL, with 2.0+ it should work fine.

domcleal:
Oh okay, I wouldn't worry about it, it isn't going to change!

PR #122, opened 2015-09-18
Dump default parameter values using --noop

domcleal:
When running Puppet to dump default parameter values, use the --noop
flag as a backup measure to ensure that resources in "params" classes
have no way of making actual system state changes.  The dump_values
parser function is unaffected.

---

https://github.com/theforeman/puppet-foreman_proxy/pull/192 is the module issue that has been fixed, but this will act as a safety net in case it happens again.  Run `foreman-installer -v` on nightly now to see the bug where a change happens while dumping default values.


ares:
Thanks @domcleal, merging!

PR #123, opened 2015-12-03
Refs #10162 - adds ability to diff configurations

mbacovsky:
- refactoring
- moved configuration specific logic from KafoConfigure to Configuration object
- ability to have multiple Configuration instances
- configuration merging
- configuration diff printing
- documentation

Tests are in progress and will be added in separate commit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,7 +221,7 @@ to look for module parameters. There are multiple ways how the installer can sel
 ```
   * by user selection in interractive mode (`-i` or `--interractive`)
 ```bash
-  $> foreman-installer -i
+  foreman-installer -i
 
   Select installation scenario
 
@@ -236,6 +236,43 @@ to look for module parameters. There are multiple ways how the installer can sel
   * automatically if there is only one scenario available
   * automatically if installer was ran already with scenario selected
 
+### Re-installing with different scenario
+
+Lets assume you have already completed installation with one scenario (e.g. smart-proxy).
+Now you want to reinstall or upgrade with different scenario (e.g. foreman). This is tricky
+situation and may end with unpredictable results so you should double check
+if the scenario and the puppet modules used in it support such kind of change.
+
+Installer tries to prevent unintentional change of a scenario and interrupts when such situation is detected:
+```bash
+  foreman-installer -S foreman-installer
+  ERROR: You are trying to replace existing installation with different scenario. This may lead to unpredictable states. Use --force to override. You can use --compare-scenarios to see the differences
+```
+
+To avoid losing some configuration values installer can detect differences between answer files of the two scenarios.
+To display them use either interactive mode (`-i`) or `--compare-scenarios` flag:
+```bash
+  foreman-installer --compare-scenarios --scenario foreman
+  Scenarios are being compared, that may take a while...
+
+  Values from previous installation that will be added by installer:
+    foreman_proxy::http_port: 8000 -> 8080
+
+  Values from previous installation that will be lost by scenario change:
+    foreman_proxy::plugin::abrt::enabled: true
+    ...
+```
+
+It may take some time as the installer has to evaluate default values for both scenarios. As a result it prints two lists.
+ - __Values from previous installation that will be added by installer:__ - in this list are options present in both scenarios but having different default values.
+   The only item from the example says that the default value for the new scenario is '8000' while the value for currently intalled scenario is '8080'.
+   When the new scenario is used the installer tries to keep the customized values from current installation and thus will use the `8080` value
+ - __Values from previous installation that will be lost by scenario change:__ - this list contains options that are part of current installation
+   and are missing from the new scenario. Most of the items are options from puppet modules that are disabled in the new scenario by default but were enabled
+   in the old one.
+
+If you are sure you want to proceed use `--force` to run the installation. Installer will replace
+the default values with values from the previous installation where possible as was indicated in the `--compare-scenario` output.
 
 ### Adding scenario
 
@@ -748,7 +785,7 @@ paths. In order to do that you can use following configuration options:
 Answer file is obvious. The "installer_dir" is the place where your installer is
 located. E.g. system checks will be loaded from here (under checks
 subdirectory) if not set elsewhere by `check_dirs`. You can optionally change foreman-installer modules dir
-using `modules_dir` option and hooks dir using `hook_dirs` option. `module_dirs`, `hook_dirs` and `check_dirs`
+using `module_dirs` option and hooks dir using `hook_dirs` option. `module_dirs`, `hook_dirs` and `check_dirs`
 can hold multiple directories where to look for the resources.
 
 On debian systems you may want to specify kafo modules dir
--- a/lib/kafo/configuration.rb
+++ b/lib/kafo/configuration.rb
@@ -8,10 +8,6 @@ module Kafo
   class Configuration
     attr_reader :config_file, :answer_file
 
-    def self.colors_possible?
-      !`which tput 2> /dev/null`.empty? && `tput colors`.to_i > 0
-    end
-
     DEFAULT = {
         :name                 => '',
         :description          => '',
@@ -24,7 +20,7 @@ module Kafo
         :installer_dir        => '.',
         :module_dirs          => ['./modules'],
         :default_values_dir   => '/tmp',
-        :colors               => Configuration.colors_possible?,
+        :colors               => Kafo::ColorScheme.colors_possible?,
         :color_of_background  => :dark,
         :hook_dirs            => [],
         :custom               => {},
@@ -39,7 +35,7 @@ module Kafo
 
       @answer_file = app[:answer_file]
       begin
-        @data = YAML.load_file(@answer_file)
+        @data = load_yaml_file(@answer_file)
       rescue Errno::ENOENT => e
         puts "No answer file at #{@answer_file} found, can not continue"
         KafoConfigure.exit(:no_answer_file)
@@ -64,7 +60,7 @@ module Kafo
     def app
       @app ||= begin
         begin
-          configuration = YAML.load_file(@config_file)
+          configuration = load_yaml_file(@config_file)
         rescue => e
           configuration = {}
         end
@@ -84,11 +80,31 @@ module Kafo
     end
 
     def modules
-      @modules ||= @data.keys.map { |mod| PuppetModule.new(mod).parse }.sort
+      @modules ||= @data.keys.map { |mod| PuppetModule.new(mod, KafoParsers::PuppetModuleParser, self).parse }.sort
+    end
+
+    def root_dir
+      File.expand_path(app[:installer_dir])
+    end
+
+    def check_dirs
+      [app[:check_dirs] || File.join(root_dir, 'checks')].flatten
+    end
+
+    def module_dirs
+      [app[:modules_dir] || app[:module_dirs] || (app[:installer_dir] + '/modules')].flatten.map { |dir| File.expand_path(dir) }
+    end
+
+    def gem_root
+      File.join(File.dirname(__FILE__), '../../')
+    end
+
+    def kafo_modules_dir
+      app[:kafo_modules_dir] || (gem_root + '/modules')
     end
 
     def add_module(name)
-      mod = PuppetModule.new(name).parse
+      mod = PuppetModule.new(name, KafoParsers::PuppetModuleParser, self).parse
       unless modules.map(&:name).include?(mod.name)
         mod.enable
         @modules << mod
@@ -106,7 +122,7 @@ module Kafo
         temp_dir = Dir.mktmpdir(nil, app[:default_values_dir])
         KafoConfigure.exit_handler.register_cleanup_path temp_dir
         @logger.info 'Loading default values from puppet modules...'
-        command = PuppetCommand.new("$temp_dir=\"#{temp_dir}\" #{includes} dump_values(#{params_to_dump})", ['--noop']).append('2>&1').command
+        command = PuppetCommand.new("$temp_dir=\"#{temp_dir}\" #{includes} dump_values(#{params_to_dump})", ['--noop'], self).append('2>&1').command
         result = `#{command}`
         @logger.debug result
         unless $?.exitstatus == 0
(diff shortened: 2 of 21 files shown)


mbacovsky:
- small refactoring and typo fixes
- tests for scenario diffs added

lib/kafo/scenario_manager.rb
```diff
@@ -29,17 +29,18 @@ def available_scenarios
     end
 
     def list_available_scenarios
-      say "Available scenarios"
+      say ::HighLine.color("Available scenarios", :cyan)
```
ares: please use [custom defined colors](https://github.com/theforeman/kafo/blob/master/lib/kafo/color_scheme.rb#L34) in the whole PR, the reason is that colors are changed according to color scheme (black vs white background)

feel free to add new key to color hash when needed...

lib/kafo/scenario_manager.rb
```diff
@@ -98,23 +116,50 @@ def scenario_from_args
       end
     end
 
+    def show_scenario_diff(prev_scenario, new_scenario)
+      say ::HighLine.color("Scenarios are being compared, that may take a while...", :cyan)
```
ares: colors (see above)

lib/kafo/scenario_manager.rb
```diff
@@ -29,17 +29,18 @@ def available_scenarios
...
 
     def list_available_scenarios
-      say "Available scenarios"
+      say ::HighLine.color("Available scenarios", :cyan)
       available_scenarios.each do |config_file, content|
         scenario = File.basename(config_file, '.yaml')
-        say "  #{content[:name]} (use: --scenario #{scenario})"
-        say "        " + content[:description] unless (content[:description].nil? || content[:description].empty?)
+        use = (config_file == @previous_scenario ? 'INSTALLED' : "use: --scenario #{scenario}")
+        say ::HighLine.color("  #{content[:name]} ", :yellow)
```
ares: colors (see above)

lib/kafo/scenario_manager.rb
```diff
@@ -98,23 +116,50 @@ def scenario_from_args
...
+          fail_now(message, :scenario_error)
         end
+      end
+    end
+
+    def print_scenario_diff(prev_conf, new_conf)
+      missing = new_conf.params_missing(prev_conf)
+      changed = new_conf.params_changed(prev_conf)
+
+      say ::HighLine.color("\nValues from previous installation that will be added by installer:", :yellow)
```
ares: colors (see above)

lib/kafo/scenario_manager.rb
```diff
@@ -98,23 +116,50 @@ def scenario_from_args
...
+      if changed.empty?
+        say "  No values will be updated from previous scenario"
       else
-        if ARGV.include?('--force')
-          message = "You are trying to replace existing installation with different scenario. This may lead to unpredictable states. Use --force to override."
-          fail_now(message,:scenario_error)
-        end
+        changed.each { |param| say "  #{param.module.class_name}::#{param.name}: #{param.value} -> #{prev_conf.param(param.module.class_name, param.name).value}" }
+      end
+      say ::HighLine.color("\nValues from previous installation that will be lost by scenario change:", :yellow)
```
ares: colors (see above)

lib/kafo/scenario_manager.rb
ares: please avoid adding \n as part of HighLine styling, it would cause problems if we set some style to use background color

test/kafo/configuration_test.rb
```diff
@@ -0,0 +1,73 @@
...
+    specify { File.exist?(File.expand_path(basic_config.gem_root)) }
+    specify { File.expand_path(basic_config.kafo_modules_dir).must_match %r|/modules$|}
+    specify { basic_config.temp_config_file.must_match %r|/tmp/kafo_answers_\d+.yaml|}
+
+    describe '#module_dirs' do
+      it 'takes modules_dir' do
+        cfg = { :modules_dir => './my_modules', :answer_file => 'test/fixtures/basic_answers.yaml'}
+        config_file = ConfigFileFactory.build('modules_dir', cfg.to_yaml).path
+        config = Kafo::Configuration.new(config_file, false)
+        assert_equal [File.join(current_dir, 'my_modules' )], config.module_dirs
```
ares: nit, extra space before `)`

README.md
```diff
@@ -236,6 +236,43 @@ to look for module parameters. There are multiple ways how the installer can sel
...
+### Re-installing with different scenario
+
+Lets assume you have already completed installation with one scenario (e.g. smart-proxy).
+Now you want to reinstall or upgrade with different scenario (e.g. foreman). This is tricky
+situation and may end with unpredictable results so you should double check
+if the scenario and the puppet modules used in it support such kind of change.
+
+Installer tries to prevent unintentional change of a scenario and interrupts when such situation is detected:
+```bash
+  $> foreman-installer -S foreman-installer
```
ares: other examples are either with `$` or without anything, please choose one to not introduce even more inconsistency

README.md
```diff
@@ -236,6 +236,43 @@ to look for module parameters. There are multiple ways how the installer can sel
...
+  ERROR: You are trying to replace existing installation with different scenario. This may lead to unpredictable states. Use --force to override. You can use --compare-scenarios to see the differences
+```
+
+To avoid losing some configuration values installer can detect differences between answer files of the two scenarios.
+To display them use either interactive mode (`-i`) or `--compare-scenarios` flag:
+```bash
+  $> foreman-installer --compare-scenarios --scenario foreman
```
ares: same as above

lib/kafo/scenario_manager.rb
```diff
@@ -29,17 +29,18 @@ def available_scenarios
...
-      say "Available scenarios"
+      say ::HighLine.color("Available scenarios", :cyan)
       available_scenarios.each do |config_file, content|
         scenario = File.basename(config_file, '.yaml')
-        say "  #{content[:name]} (use: --scenario #{scenario})"
-        say "        " + content[:description] unless (content[:description].nil? || content[:description].empty?)
+        use = (config_file == @previous_scenario ? 'INSTALLED' : "use: --scenario #{scenario}")
+        say ::HighLine.color("  #{content[:name]} ", :yellow)
+        say "(#{use})"
+        say "        " + content[:description] if content[:description]
```
ares: this break the listing when description is empty string
kafo

lib/kafo/scenario_manager.rb
```diff
@@ -29,17 +29,18 @@ def available_scenarios
     end
 
     def list_available_scenarios
-      say "Available scenarios"
+      say ::HighLine.color("Available scenarios", :cyan)
       available_scenarios.each do |config_file, content|
         scenario = File.basename(config_file, '.yaml')
-        say "  #{content[:name]} (use: --scenario #{scenario})"
-        say "        " + content[:description] unless (content[:description].nil? || content[:description].empty?)
+        use = (config_file == @previous_scenario ? 'INSTALLED' : "use: --scenario #{scenario}")
```
ares: detection does not work when relative path is used, I guess `File.expand_path` should be used

ares:
when comparing scenarios, I'm getting the trace 

```
/home/ares/.rbenv/versions/2.0.0-p247/gemsets/kafo1/gems/kafo-0.6.13/lib/kafo/params/password.rb:54:in `phrase': undefined method `app' for nil:NilClass (NoMethodError)
```

looking at https://github.com/theforeman/kafo/blob/master/lib/kafo/params/password.rb#L54 - we need to get configuration instance to the param

ares:
Will this change require a change in https://github.com/theforeman/foreman-installer/blob/42521b79e4e4a7a05dbf30b6fd5a2b783fb2e871/bin/foreman-installer#L21 ? I suppose that should be ok, since it's after the run so the config would be associated already...

lib/kafo/scenario_manager.rb
```diff
@@ -98,23 +116,50 @@ def scenario_from_args
...
       else
-        if ARGV.include?('--force')
-          message = "You are trying to replace existing installation with different scenario. This may lead to unpredictable states. Use --force to override."
-          fail_now(message,:scenario_error)
-        end
+        changed.each { |param| say "  #{param.module.class_name}::#{param.name}: #{param.value} -> #{prev_conf.param(param.module.class_name, param.name).value}" }
+      end
+      say ::HighLine.color("\nValues from previous installation that will be lost by scenario change:", :yellow)
+      if missing.empty?
+        say " No values from previous installation will be lost"
```
ares: one space indentation, while few lines above you used 2

lib/kafo/scenario_manager.rb
```diff
@@ -98,23 +116,50 @@ def scenario_from_args
...
+
+      say ::HighLine.color("\nValues from previous installation that will be added by installer:", :yellow)
+      if changed.empty?
+        say "  No values will be updated from previous scenario"
       else
-        if ARGV.include?('--force')
-          message = "You are trying to replace existing installation with different scenario. This may lead to unpredictable states. Use --force to override."
-          fail_now(message,:scenario_error)
-        end
+        changed.each { |param| say "  #{param.module.class_name}::#{param.name}: #{param.value} -> #{prev_conf.param(param.module.class_name, param.name).value}" }
```
ares: I'd expect opposite order $previous -> $new_value

lib/kafo/configuration.rb
```diff
@@ -151,17 +171,47 @@ def params
...
+      params.select do |par|
+        old_param = old_config.param(par.module.class_name, par.name)
+        old_param && old_param.value != par.value
+      end
+    end
+
+    def params_missing(old_config)
+      # finds params that are present but will be missing in the new config
+      old_config.params.select do |par|
+        param(par.module.class_name, par.name).nil?
```
ares: what would be the example here? if it's the same module, the parameter is always there and gets the default value

ares:
suppose in my old scenario answer file I have following 
```
foreman: false
```

while in new
```
foreman:
  foreman_url: something
```

I'm getting `foreman::foreman_url: ["something"] -> ["https://edna.brq.redhat.com"]` which is probably not expected. I guess we should check whether module is enabled (https://github.com/theforeman/kafo/blob/master/lib/kafo/puppet_module.rb#L36)

also the opposite situation gives me `foreman::foreman_url: ["https://edna.brq.redhat.com"] -> ["something"]`

lib/kafo/scenario_manager.rb
mbacovsky: I agree this is confusing. Currently it states that:
$new_default_value will be replaced with $previous as it may be custom value, Thus '$new -> $previous' made better sense to me. How about changing the description to: `Defaults that will be updated with values from previous installation`?

lib/kafo/configuration.rb
mbacovsky: This will cover situation, when module present in previous is missing from new scenario. All affected parameters from previous will be listed. E.g. I had proxy installed and wanted to reinstall full foreman scenario. There were some proxy plugins missing from the new scenario and this warned me.

mbacovsky:
I've updated the PR, testing it.
- I need to experiment with the modules turned off yet
- I'm not aware of reason for changes in `foreman-installer`, do you have any specific concerns?
- still not sure about `$previous -> $new_value` check my comment above

lib/kafo/scenario_manager.rb
```diff
@@ -29,17 +29,18 @@ def available_scenarios
...
-      say "Available scenarios"
+      say ::HighLine.color("Available scenarios", :info)
       available_scenarios.each do |config_file, content|
         scenario = File.basename(config_file, '.yaml')
-        say "  #{content[:name]} (use: --scenario #{scenario})"
-        say "        " + content[:description] unless (content[:description].nil? || content[:description].empty?)
+        use = (File.expand_path(config_file) == @previous_scenario ? 'INSTALLED' : "use: --scenario #{scenario}")
+        say ::HighLine.color("  #{content[:name]} ", :title)
+        say "(#{use})"
+        say "        " + content[:description] unless content[:description].nil? || content[:description].empty?
```
ares: please don't combine `unless` and `||`, I'd prefer `if` with two negations

lib/kafo/kafo_configure.rb
```diff
@@ -39,6 +40,7 @@ def hooking
     end
 
     def initialize(*args)
+      self.class.preset_color_scheme
```
ares: since this happens before the logger is built, could you make the color scheme used at https://github.com/theforeman/kafo/blob/master/lib/kafo/logger.rb#L95 so we don't duplicate ARGV parsing logic?

lib/kafo/scenario_manager.rb
ares: yeah that sounds better

lib/kafo/configuration.rb
ares: ok, that make sense, e.g. template plugin can be missing when proxy is installed with foreman, thanks for clarification

mbacovsky:
@ares: I fixed things from your comments. I also updated the way how params are diffed and took module "enabledness" into account. Check the comments in the last two commit logs for more details.

ares:
except one nit looks good to me, could you squash? I'd test and merge accordingly tomorrow

mbacovsky:
@ares, thanks for prompt review. Squashed, nit fixed.

lib/kafo/scenario_manager.rb
```diff
@@ -98,23 +116,62 @@ def scenario_from_args
...
+    end
+
+    def print_scenario_diff(prev_conf, new_conf)
+      missing = new_conf.params_missing(prev_conf)
+      changed = new_conf.params_changed(prev_conf)
+
+      say "\n" + ::HighLine.color("Overview of modules used in the scenarios (#{prev_conf.app[:name]} -> #{new_conf.app[:name]}):", :title)
+      modules = Hash.new { |h, k| h[k] = {} }
+      modules = prev_conf.modules.inject(modules) { |mods, mod| mods[mod.name][:prev] = mod.enabled?; mods }
+      modules = new_conf.modules.inject(modules) { |mods, mod| mods[mod.name][:new] = mod.enabled?; mods }
```
ares: `:prev` and `:new` is much better :-)

ares:
Thanks @mbacovsky, merging!

mbacovsky:
Thanks @ares, very much appreciated :beers:!

PR #125, opened 2015-12-23
Remove RPM spec file and accessories

domcleal:
Moved to foreman-packaging repository to match other packages and
Debian packages.

Paired with https://github.com/theforeman/foreman-packaging/pull/938


ares:
Thanks @domcleal, merging!

PR #126, opened 2016-01-18
Fixes #13046 - add pre_migrations hook

mbacovsky:
- added `pre_migration` hook which allow us to fiddle with installer config and reload it to get the changes in 
- moved confirmation of scenario change to later stage (after the migrations)
- added flush of logs even if kafo ends before logger setup. 
- fixed minor issues in logging setup when verbose is used

### DIFF
--- a/lib/kafo/hooking.rb
+++ b/lib/kafo/hooking.rb
@@ -2,6 +2,7 @@ require 'kafo/hook_context'
 
 module Kafo
   class Hooking
+    # pre_migrations - just after kafo reads its configuration - useful for config file updates. Only in this stage it is posible to request config reload (`Kafo.request_config_reload`) to get in our changes
     # boot - before kafo is ready to work, useful for adding new app arguments, logger won't work yet
     # init - just after hooking is initialized and kafo is configured, parameters have no values yet
     # pre_values - just before value from CLI is set to parameters (they already have default values)
@@ -9,7 +10,7 @@ module Kafo
     # pre_commit - after validations or interactive wizard have completed, all parameter values are set but not yet stored in the answer file
     # pre - just before puppet is executed to converge system
     # post - just after puppet is executed to converge system
-    TYPES = [:boot, :init, :pre, :post, :pre_values, :pre_validations, :pre_commit]
+    TYPES = [:pre_migrations, :boot, :init, :pre, :post, :pre_values, :pre_validations, :pre_commit]
 
     attr_accessor :hooks, :kafo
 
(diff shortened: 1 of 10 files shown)


lib/kafo/hooking.rb
```diff
@@ -9,7 +9,7 @@ class Hooking
     # pre_commit - after validations or interactive wizard have completed, all parameter values are set but not yet stored in the answer file
     # pre - just before puppet is executed to converge system
     # post - just after puppet is executed to converge system
-    TYPES = [:boot, :init, :pre, :post, :pre_values, :pre_validations, :pre_commit]
+    TYPES = [:pre_migrations, :boot, :init, :pre, :post, :pre_values, :pre_validations, :pre_commit]
```
ares: pls document above

lib/kafo/kafo_configure.rb
```diff
@@ -43,30 +43,45 @@ def initialize(*args)
...
-      self.class.scenario_manager.list_available_scenarios if ARGV.include?('--list-scenarios')
+      scenario_manager.list_available_scenarios if ARGV.include?('--list-scenarios')
+      setup_config(config_file)
+
+      self.class.hooking.execute(:pre_migrations)
+      # run migrations
+
+      # reload config
+      if @config_reload_requested
+        scenario_manager = setup_scenario_manager
```
ares: why creating `scenario_manager` local var? it does not seem to be used later again, so `self.class.scenario_manager = setup_scenario_manager` should be fine

lib/kafo/scenario_manager.rb
```diff
@@ -184,12 +187,17 @@ def link_last_scenario(config_file)
...
       conf.preset_defaults_from_puppet
       conf.preset_defaults_from_yaml
       conf
     end
 
+    def load_configuration(config_file)
+      Configuration.new(config_file)
+    end
+
+
```
ares: extra line

lib/kafo/scenario_manager.rb
```diff
@@ -198,5 +206,20 @@ def fail_now(message, exit_code)
...
+        if !KafoConfigure.config.nil?
+          Logger.setup
+          KafoConfigure.logger.info("Log was be written to #{KafoConfigure.config.log_file}")
+        end
+        KafoConfigure.logger.info('Logs flushed')
+      end
+      KafoConfigure.exit(code)
+    end
+
+
```
ares: extra lines

test/kafo/configuration_test.rb
```diff
@@ -71,5 +71,27 @@ module Kafo
         end
       end
     end
+
+    describe '#migrate_configuration' do
+      it 'migrates values from other configuration' do
+        keys = [:log_dir, :log_name, :log_level, :no_prefix, :default_values_dir,
```
ares: these 3 tests could be dried up, I'd suggest adding before block to description that will set all keys like here and in each test just run the migration with different options (nothing, skip, with) and check the result

ares:
works nicely, ready to merge

ares:
Thanks @mbacovsky, merging!

PR #127, opened 2016-01-19
Fixes #10292 - Add support for scenario updates (migrations)

mbacovsky:
- Kafo::Configuration was extended vith ability to run migrations stored in installer-scenarios.d/scenario.migrations/
- each scenario has own set of migrations
- migrations are run just once
- migrations are run after 'pre-migrations' hooks
- migrations run within limited migration context

### DIFF
--- a/lib/kafo/configuration.rb
+++ b/lib/kafo/configuration.rb
@@ -240,6 +240,22 @@ module Kafo
       @data
     end
 
+    def run_migrations
+      migrations = Kafo::Migrations.new(migrations_dir)
+      @app, @data = migrations.run(app, answers)
+      if migrations.migrations.count > 0
+        @modules = nil # force the lazy loaded modules to reload next time they are used
+        save_configuration(app)
+        store(answers)
+        migrations.store_applied
+        @logger.info("#{migrations.migrations.count} migration/s were applied. Updated configuration was saved.")
+      end
+    end
+
+    def migrations_dir
+      @config_file.gsub(/\.yaml$/, '.migrations')
+    end
+
     private
 
     def custom_storage
--- a/lib/kafo/migrations.rb
+++ b/lib/kafo/migrations.rb
@@ -1,0 +1,54 @@
+require 'yaml'
+require 'kafo/migration_context'
+
+module Kafo
+  class Migrations
+
+    attr_reader :migrations
+
+    def initialize(migrations_dir)
+      @migrations_dir = migrations_dir
+      @migrations = {}
+      @applied_file = File.join(@migrations_dir, '.applied')
+      load_migrations
+    end
+
+    def applied
+      @applied ||= load_applied
+    end
+
+    def load_migrations
+      Dir.glob(@migrations_dir + "/*.rb").each do |file|
+        next if applied.include?(File.basename(file))
+        KafoConfigure.logger.debug "Loading migration #{file}"
+        migration = File.read(file)
+        migration_block = proc { instance_eval(migration, file, 1) }
+        add_migration(file, &migration_block)
+      end
+    end
+
+    def add_migration(name, &block)
+      @migrations[name] = block
+    end
+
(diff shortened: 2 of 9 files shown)


test/fixtures/basic_answers.yaml
```diff
@@ -1 +1,10 @@
...
+# Format:
+# <classname>: false - don't include this class
+# <classname>: true - include and use the defaults
+# <classname>:
+#   <param>: <value> - include and override the default(s)
+#
+# See params.pp in each class for what options are available
+
+--- 
+  puppet: true
```
ares: missing EOF

lib/kafo/migrations.rb
```diff
@@ -0,0 +1,56 @@
...
+      @migrations[name] = block
+    end
+
+    def run(scenario, answers)
+      @processed = 0
+      @migrations.keys.sort.each do |name|
+        KafoConfigure.logger.debug "Executing migration #{name}"
+        migration = @migrations[name]
+        scenario, answers = Kafo::MigrationContext.execute(scenario, answers, &migration)
+        @processed += 1
```
ares: is @processed == @migrations.size? all loaded migrations will be executed here I think

lib/kafo/configuration.rb
```diff
@@ -240,6 +240,23 @@ def answers
       @data
     end
 
+    def run_migrations
+      migrations = Kafo::Migrations.new(migrations_dir)
+      migrations.load_migrations
+      @app, @data = migrations.run(app, answers)
+      if migrations.processed > 0
+        @modules = nil
```
ares: I guess this is to reload modules, could you add short comment here if it's the case?

lib/kafo/configuration.rb
```diff
@@ -240,6 +240,23 @@ def answers
       @data
     end
 
+    def run_migrations
+      migrations = Kafo::Migrations.new(migrations_dir)
+      migrations.load_migrations
```
ares: maybe this call could be part of initializer?

ares:
a great addition, I added few small comments, once we resolve these I'll do the testing

test/fixtures/basic_answers.yaml
mbacovsky: This file is unfortunately updated during the tests. This is how config.store(@data) wrote it. I'm thinking of moving answer file to temp file so the fixture is not changed.

lib/kafo/migrations.rb
mbacovsky: `migrations.size` would be better pointer here. Applied can contain applied migrations from past. The @processed was taken in for cases when there is an error during the migrations and only part was applied. My plan was to recover and report, but I changed my mind, I'll remove it.

lib/kafo/configuration.rb
mbacovsky: You mean  the run_migrations is invoked in initializer? That would make good sense.

test/fixtures/basic_answers.yaml
ares: ah I see, but this should happen only on tests run right? you could leave it as it was

lib/kafo/configuration.rb
ares: no I meant moving `Kafo::Migrations#load_migrations` to become part of `Kafo::Migrations#initialize`, so it wouldn't have to be called explicitly here, I don't see any use of `Kafo::Migrations` that wouldn't have loaded migration already

mbacovsky: It is useful for testing, you can manually feed in the migrations.

mbacovsky:
PR updated, all the concerns were fixed. I updated the config file factory to create temporary answer files with original content, so that the fixture is kept untouched.

ares:
Works fine, merging and going to release 0.7.0! Thanks @mbacovsky.

PR #128, opened 2016-01-28
Fixes #13441 - Add migrations only switch

mbacovsky:


lib/kafo/kafo_configure.rb
```diff
@@ -266,6 +273,7 @@ def set_app_options
       self.class.app_option ['--list-scenarios'], :flag, 'List available installation scenaraios'
       self.class.app_option ['--force'], :flag, 'Force change of installation scenaraio'
       self.class.app_option ['--compare-scenarios'], :flag, 'Show changes between last used scenario and the scenario specified with -S or --scenario argument'
+      self.class.app_option ['--migrations-only'], :flag, 'Apply migrations to a selected scenario and exit'
```
ehelms: This indicates selected scenario, does that imply I need to run this with `--scenario` ? or Specify a scenario to the `--migrations-only` command?

mbacovsky: For every installer run you have to select scenario (last used, interractively or with --scenario option). I wanted to point out, that it apply the migrations to the selected scenario only.

ehelms: :+1:

ares:
Thanks @mbacovsky, merging!

PR #129, opened 2016-01-29
Fixes #13467: Allow for relative config directories with last scenario

ehelms:


ehelms:
Note this essentially results in relative symlinks -- I could change it around to support full symlinks or relative depending on the config file path itself.

ehelms:
This seems to be breaking the scenario_changed? check, let me investigate more.

mbacovsky:
@ehelms, would it work to convert config to full path either in foreman-installer bin or in ScenarioManager init?

ehelms:
I am not sure how to do that, the setup I am trying to support is where I want to test module changes and updates an run an installation. I typically do this by having the following structure:

katello-deploy/
  katello-installer

In the scenario version:

katello-deploy/
  katello-installer/
  foreman-installer/
    config/

IN that case, when it gets copied over to a VM, I could assume it will always be /vagrant/foreman-installer/config, however, that only works in that case? I would also be open to re-fining the strategy and updating scripts based upon that. Keeping in mind, I originally designed this so that I could easily make module changes across multiple modules and run installation tests easily.

ares:
@ehelms @mbacovsky what's the status here?

ares:
@ehelms do you plan to work on this in near future? I'd like to clean up stale PRs (it could be reopened later of course)

ehelms:
I haven't seen as much of a need for this with recent changes.

PR #134, opened 2016-04-01
Fix Puppet module tests that weren't being run

domcleal:
In MiniTest 4.x, there's an issue with specify blocks alongside
describe blocks at the same level, the specs inside the describe blocks
won't run. Move all tests in this level into describe blocks so they all
run.

---

https://gist.github.com/ed3fc43a89a21761cc08ef490ab0cfd0 is a simplified case if you want to look at the bug in minitest.  Version 5 works fine, but because it has random test ordering, it introduces a whole load more issues.  I've got most of them fixed on my minitest-5 branch, except for one around loading Puppet validation functions, but have now spent way too long on it...

### DIFF
--- a/test/kafo/puppet_module_test.rb
+++ b/test/kafo/puppet_module_test.rb
@@ -85,19 +85,23 @@ module Kafo
         end
       end
 
-      let(:groups) { parsed.groups.map(&:name) }
-      specify { groups.must_include('Parameters') }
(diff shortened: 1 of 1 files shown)


ares:
[test]

ares:
Thanks @domcleal, merging! Good catch.

PR #137, opened 2016-04-07
fixes #14450 - replace validation functions with internal impls

domcleal:
Currently kafo_parsers returns a list of validation functions using the
Puppet AST, and kafo loads these from the modulepath and executes them
in a mock Puppet environment.

This reimplements the common stdlib functions in kafo, removing the need
to set up and run Puppet functions themselves. The code accepts either
the existing AST objects or simpler objects from the parser that respond
to #name and #arguments, allowing this to be switched out in
kafo_parsers later (so the Puppet library isn't needed at all).

---

In the much longer term, I don't think these will be needed when Puppet 3 support is dropped.  I think we can probably adopt the Puppet 4 type system to extend the existing type support in Kafo, which was intended to mostly replace the need for validate_* functions.


domcleal:
[test]

lib/kafo/validator.rb
```diff
@@ -2,79 +2,111 @@
...
+      return true
+    rescue TypeError, ArgumentError
+      @logger.error "#{value.inspect} is not a valid integer."
+      return false
     end
 
-    # for puppet >= 3
-    def [](value, *args)
-      lookupvar(value)
+    def validate_listen_on(args)
```
ares: ideally Kafo should provide a way to add custom validations but I guess it's fine for now :-)

lib/kafo/validator.rb
```diff
@@ -2,79 +2,111 @@
...
-        @logger.debug "calling #{method.inspect} with #{args.inspect}"
-        engine.send(method, *args)
-        true
+    def validate_re(args)
+      value = args[0]
+      regexes = args[1]
+      regexes = [regexes] unless regexes.is_a?(Array)
+      message = args[2] || "#{value.inspect} does not match the accepted inputs: #{regexes.join(", ")}"
+
+      if regexes.all? { |rx| value =~ Regexp.compile(rx) }
```
ares: when args[1] is a `Puppet::Parser::AST::ASTArray` it gets wrapped by `[]` and on this line fails with `initialize': 

```
[root@installertest kafo-0.7.3]# foreman-installer --foreman-email-delivery-method wrong             
/usr/local/share/gems/gems/kafo-0.7.3/lib/kafo/validator.rb:85:in `initialize': no implicit conversion of Puppet::Parser::AST::ASTArray into String (TypeError)
        from /usr/local/share/gems/gems/kafo-0.7.3/lib/kafo/validator.rb:85:in `compile'
        from /usr/local/share/gems/gems/kafo-0.7.3/lib/kafo/validator.rb:85:in `block in validate_re'
        from /usr/local/share/gems/gems/kafo-0.7.3/lib/kafo/validator.rb:85:in `each'
        from /usr/local/share/gems/gems/kafo-0.7.3/lib/kafo/validator.rb:85:in `all?'
        from /usr/local/share/gems/gems/kafo-0.7.3/lib/kafo/validator.rb:85:in `validate_re'
        from /usr/local/share/gems/gems/kafo-0.7.3/lib/kafo/param.rb:90:in `block in valid?'
        from /usr/local/share/gems/gems/kafo-0.7.3/lib/kafo/param.rb:90:in `each'
        from /usr/local/share/gems/gems/kafo-0.7.3/lib/kafo/param.rb:90:in `all?'
        from /usr/local/share/gems/gems/kafo-0.7.3/lib/kafo/param.rb:90:in `valid?'
        from /usr/local/share/gems/gems/kafo-0.7.3/lib/kafo/kafo_configure.rb:365:in `block in validate_all'
        from /usr/local/share/gems/gems/kafo-0.7.3/lib/kafo/kafo_configure.rb:364:in `map'
        from /usr/local/share/gems/gems/kafo-0.7.3/lib/kafo/kafo_configure.rb:364:in `validate_all'
        from /usr/local/share/gems/gems/kafo-0.7.3/lib/kafo/kafo_configure.rb:135:in `execute'
        from /usr/share/gems/gems/clamp-1.0.0/lib/clamp/command.rb:68:in `run'
        from /usr/share/gems/gems/clamp-1.0.0/lib/clamp/command.rb:133:in `run'
        from /usr/local/share/gems/gems/kafo-0.7.3/lib/kafo/kafo_configure.rb:156:in `run'
        from /usr/sbin/foreman-installer:12:in `<main>'
```

to fix it I had to add check on line 82 for this class and call `Regexp.compile(rx.to_s)` on this line, since ASTArray contains ASTString which would fail with similar error

ares:
except one issue I it looks good

lib/kafo/validator.rb
domcleal: Yeah, I agree.  I'm happy to remove this if you'd like?

lib/kafo/validator.rb
domcleal: Good catch, I've updated the interpret_validation_args method to handle arrays (things that respond to #to_a) and recursively convert those arguments into regular objects.  I found my validate_re implementation was also incorrect (it checked for all?, not any?), so that's fixed too.

Lastly, I've fixed handling of the error message argument in interpret_validation_args as it was a "concat" AST object (string with interpolation), causing errors in the logger when the validation failed.

domcleal:
And back to WoC due to test failures on Ruby 1.8.7, apologies.

domcleal:
I've broken Puppet 2.6 tests due to the manifest I'm testing at https://github.com/theforeman/kafo/pull/137/files#diff-6e61ae03fe150d8848ed11475d27ae85R134. I can add a skip on the test, or we could remove 2.6 from the matrix, as we did in kafo_parsers?

ares:
No need to test on 2.6 I guess, I removed it from both jenkins jobs. [test]

lib/kafo/validator.rb
ares: no, maybe just add a comment that this is not a standard validation, it's fine until we have a better way

mmoll:
:+1: for removing Puppet 2.6 everywhere, it's really too old now :)

lib/kafo/validator.rb
domcleal: Comment added.

domcleal:
Thanks, it's passing now without 2.6 in the matrix.

ares:
Thanks @domcleal, merging!

PR #138, opened 2016-04-07
refs #14448 - use kafo_parsers .find_available to load parser

domcleal:
By using the parser cache, the parser is now optional and will raise an
error at runtime if the cache misses and no parser is available.

---

Depends on https://github.com/theforeman/kafo_parsers/pull/14, but is also based on PR #135 as I'm trying to change the behaviour when we have both no parser and a cache.  I can swap this to do the available/soft deps change followed by the cache if preferred.

### DIFF
--- a/lib/kafo/puppet_module.rb
+++ b/lib/kafo/puppet_module.rb
@@ -2,7 +2,7 @@
 require 'kafo/param'
 require 'kafo/param_builder'
 require 'kafo/parser_cache_reader'
-require 'kafo_parsers/puppet_module_parser'
+require 'kafo_parsers/parsers'
 require 'kafo/validator'
 
 module Kafo
@@ -12,7 +12,21 @@ module Kafo
     attr_reader :name, :identifier, :params, :dir_name, :class_name, :manifest_name, :manifest_path,
                 :groups, :params_path, :params_class_name, :configuration, :raw_data
 
-    def initialize(identifier, parser = KafoParsers::PuppetModuleParser, configuration = KafoConfigure.config)
+    def self.find_parser
+      @parser ||= begin
+        logger = KafoConfigure.logger
+        parser = KafoParsers::Parsers.find_available(:logger => logger)
+        if parser
+          logger.debug "Using Puppet module parser #{parser}"
+          parser
+        else
+          logger.debug "No available Puppet module parser found"
+          :none  # prevent continually re-checking
+        end
+      end
+    end
+
+    def initialize(identifier, parser = self.class.find_parser, configuration = KafoConfigure.config)
       @identifier        = identifier
       @configuration     = configuration
       @name              = get_name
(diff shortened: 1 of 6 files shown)


domcleal:
This will continue to fail tests until kafo_parsers is released, or I could update the jobs to load kafo_parsers from GitHub so master's always tested against master.

ares:
>  I could update the jobs to load kafo_parsers from GitHub so master's always tested against master

no that's fine, I hope to release kafo_parsers 0.1 soon, just after https://github.com/theforeman/kafo_parsers/pull/13 gets in

domcleal:
Rebased to use kafo_parsers 0.1.0.

domcleal:
I've added a cache variable into PuppetModule.find_parser, as the .available? method on the puppet-strings parser is very slow.  This caused long delays when the parser's not available, as it was checking availability for each Puppet module.

lib/kafo/puppet_module.rb
```diff
@@ -12,7 +12,11 @@ class PuppetModule
     attr_reader :name, :identifier, :params, :dir_name, :class_name, :manifest_name, :manifest_path,
                 :groups, :params_path, :params_class_name, :configuration, :raw_data
 
-    def initialize(identifier, parser = KafoParsers::PuppetModuleParser, configuration = KafoConfigure.config)
+    def self.find_parser
+      @@parser ||= KafoParsers::Parsers.find_available(:logger => KafoConfigure.logger) || :none
```
ares: why class variable? I don't see it used anywhere else, for memoization instance variable should be fine

ares:
:+1: except one nit

lib/kafo/puppet_module.rb
domcleal: This is so it's cached when creating multiple PuppetModules, e.g. an installer with a dozen modules would call find_parser once for each module, which is very slow when using the puppet-strings parser (its available? method).  An instance variable wouldn't work across multiple instances of PuppetModule.

I could move this into Kafo::Configuration#modules instead?

domcleal:
Updated to change another instance of PuppetModuleParser to use PuppetModule.find_parser in Kafo::Configuration#add_module.

lib/kafo/puppet_module.rb
ares: But this is a class method, so instance scope is PuppetModule object scope. The difference is when some class would inherit from PuppetModule, so the class variable would be shared with children which I don't think is expected.

```ruby
class A
  def self.a
    @a ||= rand(1000)
  end
end

10.times {
  puts A.new.class.a # will always print the same number
}
```

domcleal: Ah yes, thanks!  Fixed, changed to `@parser`.

lib/kafo/puppet_module.rb
```diff
@@ -49,9 +53,15 @@ def enable
...
     def parse(builder_klass = ParamBuilder)
-      @params      = []
-      @raw_data    = @parser_cache.get(identifier, manifest_path) if @parser_cache
-      @raw_data  ||= @parser.parse(manifest_path)
+      @raw_data = @parser_cache.get(identifier, manifest_path) if @parser_cache
+      if @raw_data.nil?
+        if @parser.nil? || @parser == :none
+          raise ParserError.new("No Puppet module parser is installed and no cache of the file #{manifest_path} is available. Please check debug logs and install optional dependencies for the parser.")
+        else
+          @raw_data = @parser.parse(manifest_path)
```
ares: During testing I realized that it would be great if it logged out a debug message, which `@parser.class` which is being used. Maybe a better place would be `self.find_parser` so it would be printed once rather than for every module.

domcleal: Yep, good idea, that'd be very useful.  Updated find_parser to include debug statements.

ares:
Thanks @domcleal, I'll bump the version to 0.8 once #137 gets in. I can also add more puppet versions to Jenkins when we have 0.8.0 build.

domcleal:
> I can also add more puppet versions to Jenkins when we have 0.8.0 build.

I'm unsure if this will work well at the moment because there are some tests that depend on the PuppetModuleParser to return lists of validations, which is why I kept the `< 4.0` development dep.  There might be other reasons, or it might be simpler than I thought - I've not looked much at it yet.

PR #139, opened 2016-04-08
fixes #14452 - search /opt/puppetlabs/bin to run Puppet

domcleal:
Preferring to search rather than just set PATH for improved logging, since the user will see in logs precisely which Puppet binary (if multiple exist) was used to run the command, which may help us in debugging.

### DIFF
--- a/lib/kafo/puppet_command.rb
+++ b/lib/kafo/puppet_command.rb
@@ -25,7 +25,7 @@ module Kafo
           "echo '$kafo_config_file=\"#{@configuration.config_file}\" #{custom_answer_file} #{add_progress} #{@command}'",
           '|',
           "RUBYLIB=#{["#{@configuration.gem_root}/modules", ::ENV['RUBYLIB']].join(File::PATH_SEPARATOR)}",
-          "puppet apply #{@options.join(' ')} #{@suffix}",
+          "#{puppet_path} apply #{@options.join(' ')} #{@suffix}",
       ].join(' ')
       @logger.debug result
       result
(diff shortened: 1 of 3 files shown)


ekohl:
I like the explicitness of this, but I wonder if we should prefer AIO over $PATH.

domcleal:
If we add the AIO path to the start of the $PATH then the user's left with no way to change the behaviour unless we added flags to Kafo itself.  At least if PATH is preferred and there's another `puppet` in the PATH, the user could always add /opt/puppetlabs/bin to the start of their PATH to override which is used.

ekohl:
You're right, that makes sense. We should document that option somewhere.

domcleal:
Good idea, I added a section to the README on how Puppet binaries are found.

ekohl:
:+1:

lib/kafo/puppet_command.rb
```diff
@@ -44,5 +44,13 @@ def modules_path
           @configuration.kafo_modules_dir,
       ].flatten.join(':')
     end
+
+    def puppet_path
+      (::ENV['PATH'].split(File::PATH_SEPARATOR) + ['/opt/puppetlabs/bin']).find do |path|
```
ares: I know it's unlikely that another puppet would be installed during default values parsing but I think this should be cached so that the same puppet binary is called for both puppet runs, an example can be found at https://github.com/theforeman/kafo_parsers/blob/792d0096741d830966e9456b44f5d8891f6c2e16/lib/kafo_parsers/puppet_strings_module_parser.rb#L57

since `#find` is lazy, the return was not necessary here if you use `|| 'puppet'`

maybe something like this would work

```ruby
def puppet_path
  @puppet_path ||= begin
    bin_name = 'puppet'
      File.executable?(File.join(path, bin_name))
    end
    File.join([bin_path, bin_name].compact)
  end
end
```

domcleal: Thanks, updated to use your suggested code, it makes sense.

ares:
Works just fine, thank you @domcleal, merging!

PR #140, opened 2016-04-11
fixes #14568 - don't pass any default value for undef class parameters

domcleal:
Under Puppet 4, dump_values returned the parameter name for any undef
values, as lookupvar now returned nil rather than :undef. This has been
removed so it simply returns nil in all cases.

Secondly, when instantiating classes with undef parameter values, don't
pass the parameter at all when it's nil (undef) to allow it to be
determined by Puppet. Passing an undef value causes the parameter to be
set to the value undef, rather than meaning "not set".


ares:
Works fine, thank you @domcleal, merging!

PR #141, opened 2016-04-11
Fixes #14571 - expand multiple module dirs

ares:


bin/kafo-export-params
```diff
@@ -31,7 +31,7 @@ module Kafo
       c                         = Configuration.new(config, false)
       KafoConfigure.config      = c
       KafoConfigure.root_dir    = File.expand_path(c.app[:installer_dir])
-      KafoConfigure.module_dirs = File.expand_path(c.app[:module_dirs])
+      KafoConfigure.module_dirs = c.app[:module_dirs].map { |path| File.expand_path(path) }
```
domcleal: If module_dirs is set to a string (the default with config/kafo.yaml.example), this throws an error as the app config's simply a string.  Configuration uses `[app[:module_dirs]].flatten`, perhaps adopt this?  Or even better, can you just call `c.module_dirs` here?

ares: right, Configuration class does expanding, I missed that

domcleal:
Thanks @ares!

PR #142, opened 2016-04-14
Fixes #14657: Check output of each directories system checker output

ehelms:


test/fixtures/checks/pass/pass.sh
```diff
@@ -0,0 +1 @@
+exit 0
```
domcleal: Missing a shebang line, this seems to fail under Ruby 1.8.7.

test/fixtures/checks/fail/fail.sh
```diff
@@ -0,0 +1 @@
+exit 1
```
domcleal: Should have a shebang line as well, to be safe.

domcleal:
:+1: to the fix, it works correctly.  #143 will fix the other tests on 1.8.7.

ehelms:
Shebangs added

domcleal:
Thanks, the new tests are passing now.  Can be merged once master's passing.

ares:
[test]

ares:
Thanks @ehelms, merging!

PR #143, opened 2016-04-15
fixes #14658 - don't depend on Symbol#<=> in tests

domcleal:


ares:
Thanks @domcleal, merging.

PR #144, opened 2016-04-18
fixes #14683 - remove default parser cache location

domcleal:


ares:
I guess this would be worth of bringing back to 0.7 branch

domcleal:
Yep, if you're releasing another 0.7 then it should be included.

ares:
Thanks @domcleal, merging.

ares:
cherry-picked to 0.7 branch as 36a99cd02841bdbe90e938a3a033f5869146ac5a, if there is another 0.7 release it will be included

PR #145, opened 2016-04-21
Fixes #14758 - optimize progress bar for narrow terminal

mbacovsky:
On terminal that has width < 83 chars the progress bar is reduced to just `[]` and tend to leave orphan `[` at the end of line. This patch removes progress bar for lines < 83 chars and leaves just the percent done in the output. This way it fits well to the standard 80x25 sized term.


domcleal:
:+1: works well.

ares:
Thanks @mbacovsky, merging.

PR #149, opened 2016-05-05
fixes #14937 - interpolate $param in validation args

domcleal:
kafo_parsers 0.1.1 concatenates parsed strings itself, so variables in
validation argument strings are no longer separate AST objects and now
need to be replaced with gsub. Other AST remnants are removed to depend
on 0.1.1.

Remaining test failures will be fixed by https://github.com/theforeman/kafo_parsers/pull/22.

### DIFF
--- a/lib/kafo/param.rb
+++ b/lib/kafo/param.rb
@@ -120,12 +120,8 @@ module Kafo
       args.map do |arg|
         if arg.to_s == "$#{self.name}"
           self.value
-        elsif arg.class.name == 'Puppet::Parser::AST::ASTArray'
-          interpret_validation_args(arg.to_a)
-        elsif arg.class.name == 'Puppet::Parser::AST::Concat'
-          interpret_validation_args(arg.value).join
-        elsif arg.respond_to? :value
-          arg.value
+        elsif arg.is_a? String
+          arg.gsub("$#{self.name}", self.value.to_s)
         else
           arg
         end
--- a/kafo.gemspec
+++ b/kafo.gemspec
@@ -26,7 +26,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'kafo_wizards'
   spec.add_dependency 'ansi'
   # puppet module parsing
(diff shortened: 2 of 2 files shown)


domcleal:
[test]

lib/kafo/param.rb
```diff
@@ -120,12 +120,8 @@ def interpret_validation_args(args)
...
         if arg.to_s == "$#{self.name}"
           self.value
-        elsif arg.class.name == 'Puppet::Parser::AST::ASTArray'
-          interpret_validation_args(arg.to_a)
-        elsif arg.class.name == 'Puppet::Parser::AST::Concat'
-          interpret_validation_args(arg.value).join
-        elsif arg.respond_to? :value
-          arg.value
+        elsif arg.is_a? String
+          arg.gsub("$#{self.name}", self.value.to_s)
```
ares: I'm struggling to understand why it's not caught by the first if condition, what does the string contains as extra characters? and wouldn't be `sub` instead of `gsub` enough?

domcleal: An example validation that would trigger both parts is:

    validate_re($db_type, ["^mysql$", "^sqlite$"], "invalid $db_type DB type")

The first argument is passed back from kafo_parsers as `"$db_type"` and would be replaced by the first condition.

The third argument is passed back from kafo_parsers as `"invalid $db_type DB type"`, so is matched and replaced with gsub by this condition.

I could probably remove the first condition entirely now.

ares: oh, makes sense, so please remove the first condition

domcleal: > I could probably remove the first condition entirely now.

It might be safest to leave it, since this condition would change the argument value to a string, while the first condition retains the original data type.

ares: that's true, ok let's leave it for now, thanks!

ares:
Merged, thanks @domcleal!

PR #150, opened 2016-05-06
fixes #14941 - set Puppet RUBYLIB from kafo_modules_dir config

domcleal:
The patch at https://github.com/theforeman/foreman-packaging/blob/deb/1.11/dependencies/jessie/kafo/patches/fix_require_in_add_progress.diff can be deleted when this is released.


ares:
[test]

ares:
Thanks @domcleal, merging!

PR #151, opened 2016-05-06
fixes #14945 - use Hiera to load Puppet classes and parameters

domcleal:
Replaces use of parser functions to load and transform the answers file
by including classes with hiera_include() and using Hiera with data
binding to pass class parameters in.

This requires Puppet 3 for both hiera_include() and data binding
support.

It allows a custom Hiera config file to be set by `:hiera_config`, and
by adding `kafo_answers` to the hierarchy, parameters can be overridden
or set in a way decided by the installer developer. This will be useful
to set parameters of classes _not_ managed by Kafo, that are indirectly
included.


ares:
[test]

beav:
@domcleal I'm not very familiar with kafo dev setups but would like to try out this PR. What is the best way to install it locally to test with the katello install scenario?

domcleal:
https://github.com/theforeman/kafo#how-do-i-use-it explains how to run kafo, just add it to an installer project's Gemfile.

domcleal:
There's a ticket that's on the Foreman 1.12 issues list which depends on the functionality in this PR, so a review soon would be appreciated. I could split it up into support for a custom hiera_config and a separate change to use it for Kafo's parameters if that's more palatable.

mmoll:
would this open the possibility to solve https://github.com/theforeman/puppet-puppet/pull/361 by using puppetlabs-puppetdb 5.x and set that parameter via hiera, depending on the puppet version?

ares:
> so a review soon would be appreciated

Sorry, I will try to look at this and other PRs by end of the week.

domcleal:
> would this open the possibility to solve theforeman/puppet-puppet#361 by using puppetlabs-puppetdb 5.x and set that parameter via hiera, depending on the puppet version?

Yes, quite possibly - instead of including the globals class directly.

lib/kafo/hiera_configurer.rb
```diff
@@ -0,0 +1,76 @@
...
+      # use our copy of the data dir
+      config[:yaml] ||= {}
+      config[:yaml][:datadir] = data_dir
+
+      config
+    end
+
+    def generate_data(modules)
+      classes = []
+      data = modules.select(&:enabled?).inject({}) do |config,mod|
```
ares: nit, missing space after comma `|config, mod|`

lib/kafo/kafo_configure.rb
```diff
@@ -371,6 +369,11 @@ def validate_all(logging = true)
 
     def run_installation
       self.class.hooking.execute(:pre)
+
+      hiera = HieraConfigurer.new(config.app[:hiera_config], config.modules, config.app[:order])
+      hiera.write_configs
```
ares: Shouldn't this respect `noop?`? I suppose it would fail if `kafo_answers` source would be missing but still, noop shouldn't do any changes I think. If hiera config would be right it would work without write_configs I think.

domcleal: Everything in `#write_configs` should happen in a temporary directory - it's meant to be self-contained so we generate a set of merged Hiera configs and data files from answers from what's in memory, then we wipe it after it's used to run Puppet. No generated Hiera config or data files are kept.

Noop is still respected when saving the Kafo answers file up on lines 143-144.

README.md
```diff
@@ -158,6 +158,14 @@ As you may have noticed there are several ways how to specify arguments. Here's
   * values specified on CLI
   * interactive mode arguments
 
+## Requirements
+
+Kafo is supported with Puppet versions 3 and 4. Puppet 2 is no longer supported
```
ares: should we bump the version?

lib/kafo/kafo_configure.rb
ares: right. I was confused by `generate_config` but I see now it works under copy of original file

README.md
domcleal: Yep, I think so.

lib/kafo/hiera_configurer.rb
```diff
@@ -0,0 +1,76 @@
...
+    end
+
+    def write_configs
+      @temp_dir = Dir.mktmpdir('kafo_hiera')
+      @config_path = File.join(temp_dir, 'hiera.conf')
+      @data_dir = File.join(temp_dir, 'data')
+
+      if @user_config_path
+        logger.debug("Merging existing Hiera config file from #{@user_config_path}")
+        user_config = YAML.load(File.read(@user_config_path))
```
ares: I see you rely on Ruby interpreter seeing this variable even if it does no go through this path of condition. Could you please add user_config = {} and user_datadir = false in else branch to be explicit?

lib/kafo/hiera_configurer.rb
```diff
@@ -0,0 +1,76 @@
...
+        user_datadir = user_config[:yaml][:datadir] if user_config[:yaml]
```
ares: the name should probably be `user_data_dir` to be consistent with `@data_dir`

lib/kafo/hiera_configurer.rb
```diff
@@ -0,0 +1,76 @@
...
+      data = modules.select(&:enabled?).inject({}) do |config,mod|
+        classes << mod.class_name
+        config.update(Hash[mod.params_hash.map { |k,v| ["#{mod.class_name}::#{k}", v] }])
+      end
+      data['classes'] = @modules_order ? sort_modules(classes, @modules_order) : classes
+      data
+    end
+
+    def sort_modules(modules, order)
+      (order & modules) + (modules - order)
```
ares: :clap:

ares:
Tested with foreman-installer, works without any change. Generated files contains `!ruby/sym`. Since they are temporary it shouldn't matter too much, maybe only during debugging but that's minor. After fixing few syntax nits it's ready to be merged.

lib/kafo/hiera_configurer.rb
domcleal: Yep, both added.

lib/kafo/hiera_configurer.rb
domcleal: Renamed.

lib/kafo/hiera_configurer.rb
domcleal: Space added, and on the line below.

domcleal:
Thanks for the review, style comments addressed.  I didn't reproduce the issue with `!ruby/sym`, but I could run it through a similar search/replace as we do in the answers file?  (lib/kafo/configuration.rb format)

ares:
You could if you want, but it's really optional :-)

domcleal:
I've added the symbol replacement to both the Hiera config and data file.

ares:
Thanks @domcleal, merged. I'll release 0.9 in a moment.

PR #152, opened 2016-05-19
fixes #15071 - add --reset-* parameter options

domcleal:
Parameters can now be reset to the default value set in the manifest by
passing --reset-mod-param or by using a reset sub-menu in interactive
mode.

The behaviour of the Param class has changed to remember defaults.
Previously for a parameter whose default was $::module::params::foo,
the 'default' always remained that parameter name, Puppet was run to
fetch the value of the params class variable, then 'value' was set to
the value of the params variable.

Now 'default' is changed to the value of the params variable so we can
differentiate between the user's value and the default. When resetting
a value to the default it now goes back to the params value instead of
the params variable name.

### DIFF
--- a/README.md
+++ b/README.md
@@ -507,6 +507,25 @@ as key:value.
 When parsing the value, the first colon divides key and value. All other
 colons are ignored.
 
+## Resetting an argument
+
+Existing stored parameters can be reset back to their default value from the
+command line or interactive mode.  This deletes the stored value in the answers
+file and stores the default from the Puppet manifest in its place.
+
+The default value is the value set in, or computed by the Puppet params
+manifest. This will _not_ reset to any defaults specified in the answers file
+before running the Kafo-based installer, they are not kept.
+
+Using the CLI, a --reset option is available for every parameter, e.g.
+
+```bash
+bin/foreman-installer --reset-puppet-server-git-branch-map
+```
+
+The parameter can also be reset to the default in interactive mode, via the
+reset parameters sub-menu under each module.
+
 ## Grouping in interactive mode
 
 If your module has too many parameters you may find the grouping feature useful.
--- a/lib/kafo/help_builders/base.rb
+++ b/lib/kafo/help_builders/base.rb
@@ -82,7 +82,11 @@ module Kafo
       end
 
       def parametrization
-        @parametrization ||= Hash[@params.map { |p| [parametrize(p), p] }]
+        @parametrization ||= begin
+          @params.inject({}) do |h,p|
+            h.update(parametrize(p) => p, parametrize(p, 'reset-') => p)
+          end
+        end
       end
     end
   end
--- a/lib/kafo/param.rb
+++ b/lib/kafo/param.rb
@@ -26,6 +26,11 @@ module Kafo
       @value     = value == 'UNDEF' ? nil : value
     end
 
+    def unset_value
+      @value_set = false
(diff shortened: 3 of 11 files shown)


README.md
```diff
@@ -507,6 +507,18 @@ as key:value.
 When parsing the value, the first colon divides key and value. All other
 colons are ignored.
 
+## Resetting an argument
+
+Existing stored parameters can be reset back to their default value. This
```
ares: I'm not sure, but it might be good to better explain what's the default value. Assume that you have installer based on kafo (e.g. foreman-installer) which ships some "default" values in packaged answer file. This reset feature actually resets the value to "system default" right?

Btw, thanks for automatically documenting all new additions!

lib/kafo/help_builders/base.rb
```diff
@@ -82,7 +82,11 @@ def params_mapping(items)
       end
 
       def parametrization
-        @parametrization ||= Hash[@params.map { |p| [parametrize(p), p] }]
+        @parametrization ||= begin
+          @params.inject({}) do |h,p|
+            h.update(parametrize(p) => p, parametrize(p, 'reset-') => p)
```
ares: not tested yet, but is this necessary for each parameter? maybe we could document it somewhere just once to improve readability

domcleal: I hid all of the reset options in the basic help output (default --help) but left them in --full-help for completeness as it might show up for people who grep the output.  I do partly agree though, I was planning to hide them at first and try to add "--reset-[NAME]  Reset parameter NAME" or something instead, but didn't do that in the end.

README.md
domcleal: Yep, you're right, it's the system default implied by the manifest. I can clarify this document.

ares: ok, that would be great, thanks

domcleal: Updated to explain the default.

lib/kafo/param.rb
```diff
@@ -40,29 +45,26 @@ def to_s
...
               # newer puppet has default dump in format 'value' => 'value' so
               # it's handled correctly by else branch
-              self.value = self.default
             else
-              self.value = value
+              self.default = value
           end
           # if we don't have default value from dump (can happen for modules added from hooks)
           # we fallback to their own default values which must be sensible
-        else
```
ares: I suppose this is removed because value loads data from default unless it has been set according to @value_set, we should probably either remove the comment above as well or add one more line explaining that it will be correctly loaded from default value.

test/kafo/param_test.rb
```diff
@@ -159,5 +159,27 @@ def create_validation(function, function_args)
         validation
       end
     end
+
+    describe '#set_default' do
+      let(:with_params) { param.tap { |p| p.default = 'mod::params::test' } }
+      let(:with_unset) { param.tap { |p| p.default = 'UNSET' } }
+      let(:with_value) { param.tap { |p| p.default = 42 } }
```
ares: I guess this answers everything :-)

ares:
Tested, works well even with -d and -n. Except one nit regarding comment, it's ready to be merged.

lib/kafo/param.rb
domcleal: Thanks, I didn't notice that comment. I've updated it.

ares:
Thanks @domcleal, merging!

PR #153, opened 2016-05-20
fixes #15105 - replace list of Puppet versions with checks

domcleal:
The list of Puppet versions supporting the progress bar has been
replaced with more defensive checks when monkey patching Puppet.

### DIFF
--- a/modules/kafo_configure/lib/puppet/parser/functions/add_progress.rb
+++ b/modules/kafo_configure/lib/puppet/parser/functions/add_progress.rb
@@ -2,10 +2,16 @@ require 'kafo_configure/lib/kafo/puppet/report_wrapper'
 
 module Puppet::Parser::Functions
   newfunction(:add_progress) do |args|
-    supported = %w(2.6. 2.7. 3.0. 3.1. 3.2. 3.3. 3.4. 3.5. 3.6. 3.7. 3.8.)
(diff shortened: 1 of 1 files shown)


modules/kafo_configure/lib/puppet/parser/functions/add_progress.rb
```diff
@@ -2,10 +2,15 @@
...
-    if supported.any? { |version| Puppet::PUPPETVERSION.start_with?(version) }
-      # Monkey patch the transaction to put our wrapper around the report object
+    begin
       require 'puppet/transaction'
+      loaded = true
+    rescue LoadError
+      ::Puppet.warning 'Unable to load puppet/transaction for progress bar support, this version may not be supported'
+    end
+
+    if loaded
```
ares: I know this works even if require raises an exception since the interpreter already saw `loaded` local variable but could you please add explicit `loaded = false` on before `begin`?

ares:
except one nit looks good and works well

modules/kafo_configure/lib/puppet/parser/functions/add_progress.rb
domcleal: Sure, updated.

mmoll:
Brings back the beloved progress bar on Puppet 4 :heart:

ares:
Thanks @domcleal, merged!

PR #154, opened 2016-05-20
fixes #12166 - prevent over-counting in progress bar

domcleal:
Defined types are evaluated twice, causing two RESOURCE lines to be
printed and double-counting. A set now tracks whether a resource has
already been evaluated and prevents progress being incremented again.

Resources such as concat 2.x and recursive file (dir) resources will
use eval_generate to add more resources into the catalog during
evaluation, which isn't in the initial count. These are now ignored by
tracking which resources were initially counted and only incrementing
the progress bar when those same resources are seen during the run.

Anchor and Kafo's own yaml_to_class resources are now also ignored to
improve the accuracy of the overall resource count.

### DIFF
--- a/modules/kafo_configure/lib/kafo/puppet/report_wrapper.rb
+++ b/modules/kafo_configure/lib/kafo/puppet/report_wrapper.rb
@@ -1,3 +1,5 @@
+require 'set'
+
 module Kafo
   module Puppet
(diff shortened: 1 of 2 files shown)


ares:
big :clap: it's much more accurate now, thanks @domcleal, merging

PR #155, opened 2016-05-23
fixes #15131 - disable reports during Puppet runs

domcleal:


mmoll:
That problem probably was uncovered by https://github.com/theforeman/puppet-puppet/commit/29ec2f16b1a1f12c43c4945d95bdbfc2bf5cff64
The patch fixes it for me. :+1:

ares:
Thanks @domcleal and @mmoll, merging.

PR #156, opened 2016-05-23
fixes #15111 - replace HighLine::String in param values

domcleal:
By filtering these classes in the param setter, it's fixed for both the
wizard and users with persisted answers as the answers file is loaded
through the same setter.

### DIFF
--- a/lib/kafo/param.rb
+++ b/lib/kafo/param.rb
@@ -23,6 +23,7 @@ module Kafo
 
     def value=(value)
       @value_set = true
+      value      = value.to_s if value.is_a?(::HighLine::String)  # don't persist highline extensions
       @value     = value == 'UNDEF' ? nil : value
     end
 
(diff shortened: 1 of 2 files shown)


mmoll:
fixes the `Highline::String` problem for me :+1:

lib/kafo/param.rb
```diff
@@ -23,6 +23,7 @@ def value
 
     def value=(value)
       @value_set = true
+      value      = value.to_s if value.class == ::HighLine::String  # don't persist highline extensions
```
ares: `value.is_a?(::HighLine::String)` looks a bit more readable to me and is a bit safer since it covers all children classes (would be same as <= :HighLine::String)

ares:
one nit, otherwise good to merge

lib/kafo/param.rb
domcleal: Good suggestion, fixed.

ares:
Thanks @domcleal, merging!

PR #157, opened 2016-06-07
Fixes #15322 - reload configuration prior migrations execution

mbacovsky:
We need to reload scenario configuration after `pre_migration` hooks are finished. Otherwise the scenario migrations are run on the original config and the result replaces what hooks produced.

Second config reload is requested only when some migrations were applied


ares:
Thanks @mbacovsky, merging.

PR #160, opened 2016-06-13
Fixes #15382 - Improved error messages on missing scenario

mbacovsky:
Based on users feedback I'm trying to make the error messages more clear and instructive.
The output with the patch:
```text
# foreman-installer 
ERROR: No installation scenario was selected, the installer can not continue.
       Even --help content is dependent on selected scenario.
       Select scenario with --scenario SCENARIO or list available scenarios with --list-scenarios.
```

### DIFF
--- a/lib/kafo/logger.rb
+++ b/lib/kafo/logger.rb
@@ -105,7 +105,7 @@ module Kafo
 
     def self.dump_buffer(buffer)
       buffer.each do |log|
-        self.loggers.each { |logger| logger.send log[0], *log[1], &log[2] }
+        self.loggers.each { |logger| logger.send log[0], *([log[1]].flatten(1)), &log[2] }
       end
       buffer.clear
     end
--- a/lib/kafo/scenario_manager.rb
+++ b/lib/kafo/scenario_manager.rb
@@ -91,12 +91,12 @@ module Kafo
         (available_scenarios.keys.count == 1 && available_scenarios.keys.first) ||
         select_scenario_interactively
       if scenario.nil?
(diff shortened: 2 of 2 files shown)


ohadlevy:
@mbacovsky note it fails on ruby 1.87

lib/kafo/scenario_manager.rb
```diff
@@ -91,10 +91,13 @@ def select_scenario
         (available_scenarios.keys.count == 1 && available_scenarios.keys.first) ||
         select_scenario_interactively
       if scenario.nil?
-        fail_now("Scenario was not selected, can not continue. Use --list-scenarios to list available options.", :unknown_scenario)
+        fail_now("No installation scenario was selected, the installer can not continue.\n" \
```
ares: could you please replace trailing `\` with `+` to be more explicit?

lib/kafo/logger.rb
```diff
@@ -105,7 +105,7 @@ def dump_errors
 
     def self.dump_buffer(buffer)
       buffer.each do |log|
-        self.loggers.each { |logger| logger.send log[0], *log[1], &log[2] }
+        self.loggers.each { |logger| logger.send log[0], *([log[1]].flatten(1)), &log[2] }
```
ares: why is this needed?

mbacovsky: This was breaking tests on Ruby 1.8.7. In some of the log buffer records the messages were strings and not arrays and the old logger didn't recover from that.

ares:
Thanks @mbacovsky, merging!

PR #161, opened 2016-06-28
fixes #6978 - show validation failure reasons in output

domcleal:

PR #162, opened 2016-07-01
fixes #14500 - support Puppet 4 parameter data types

domcleal:
When kafo_parsers returns a Puppet 4 parameter type (e.g.
`Optional[Integer[10]]`), the user input to this parameter can now be
validated against the data type specification. When not supplied, it
defaults to `Any`, roughly matching the existing `string` type.

Data types have been separated out from Param and its subclasses so they
can be nested easily, as is common with `Optional`, `Array` and others,
plus types accept arguments, e.g. possible values for an enum, or max
lengths. Some of the more esoteric metatypes haven't been implemented as
it probably isn't possible to handle their values.

The existing Kafo data types are still supported by translating them to
the newer type equivalent, e.g. `string` to `Optional[String]`.

---

Based on PR #161 to save time later if that's accepted, as this way the data type validations also provide failure reasons in the same way as validation functions.


lib/kafo/data_types/boolean.rb
```diff
@@ -0,0 +1,27 @@
...
+            false
+          when '1', 'true', 't', 'y', true
+            true
+          else
+            value
+        end
+      end
+
+      def valid?(input, errors = [])
+        if input.is_a?(::TrueClass) || input.is_a?(FalseClass)
```
ares: I think this code would be more readable as
```ruby
valid = input.is_a?(::TrueClass) || input.is_a?(FalseClass)
errors << "#{input.inspect} is not a valid boolean" unless valid
valid
```

seems a bit more readable, sadly there's no #tap in 1.8.7 which would make it one-liner

also, why `::TrueClass` but `FalseClass` without absolute namespace path

lib/kafo/data_type.rb
```diff
@@ -0,0 +1,99 @@
...
+          raise ConfigurationException, "unknown data type #{keyword_re[1]}"
+        end
+      else
+        raise ConfigurationException, "data type not recognized #{str}"
+      end
+    end
+
+    def self.register_type(keyword, type)
+      @keywords ||= {}
+      @keywords[keyword] = type
```
ares: it might raise exception if the type is registered already? now when hooks can come from installer plugins this can happen more often

ares:
I have just two small comments, plus it needs rebasing. I'll test afterwards that it works fine with 3.x. I hope you're aware of this PR worsen your statistics, until now you have [removed](https://github.com/theforeman/kafo/graphs/contributors) more lines than added :-) After this is merged I'll bump the version to 0.10.

lib/kafo/data_type.rb
domcleal: Good idea, it now throws an exception on re-registration (which found a bug in the tests, now fixed).

lib/kafo/data_types/boolean.rb
domcleal: tap's available since 1.8.7 so I've rewritten it that way, and added the missing namespace `::`.  (tap wasn't in 1.8.5 on EL5)

domcleal:
That'll be all of the tests!  PR updated with the two fixes and rebased.

ares:
Thanks @domcleal, huge change means huge bump, I think it's time for 1.0.0 https://www.youtube.com/watch?v=_J6-3l3hCm0

PR #163, opened 2016-07-01
fixes #15562 - convert undef parameter defaults to nil

domcleal:
Uses the existing code to convert :undef to a nil default, but runs it
against the original parameter default for cases when a params class
isn't used.


ares:
Thank you @domcleal, tested and fixes the issue - tested with puppet 3.8.2. Merged.

PR #164, opened 2016-07-20
Fixes #15757 - check history log files if log exists

iNecas:


ares:
[test]

bbuckingham:
Changes look good to me.   In addition, tested the patch on an upgrade scenario and confirmed that answers were correctly migrated from the old/legacy answers to the new scenario-based answers file.  This was not the case prior to the patch.

ACK for me.

ares:
thanks, closing in favour of  #165

PR #166, opened 2016-08-10
refs #16230 - read a list of parser caches then merge them

komidore64:


komidore64:
[test]

komidore64:
[test]

komidore64:
ping @ares

ares:
I'll try to look at the PR by the end of the week, next time you change the name of PR to remove WIP please also write a comment that review can start, otherwise I don't get any notification.

komidore64:
alrighty, i can manage that.

:+1:

lib/kafo/parser_cache_reader.rb
```diff
@@ -1,24 +1,35 @@
...
         return nil
       end
 
-      parsed = YAML.load(File.read(cache_path))
-      if !parsed.is_a?(Hash) || parsed[:version] != 1 || !parsed[:files].is_a?(Hash)
-        logger.warn "Parser cache is from a different version of Kafo, skipping setup"
-        return nil
+      parsed = cache_paths.map { |path| YAML.load(File.read(path)) }
+
+      parsed.each_with_index do |cache, i|
```
ares: why `each_with_index`? you only use `i` later to get `cache_paths[i]` but that should be the same object as `cache` I think?

lib/kafo/parser_cache_reader.rb
```diff
@@ -1,24 +1,35 @@
...
+          return nil
+        end
       end
 
-      logger.debug "Using #{cache_path} cache with parsed modules"
-      new(parsed)
+      logger.debug "Using #{cache_paths.join(", ")} cache with parsed modules"
+
+      merged_cache = {
+        :version => 1,
```
ares: could you please extract `1` into some variable or even better constant like SUPPORTED_VERSION? Now it's on two places so if we changed it later, it would be easy to miss.

ares:
Could you please create a separate kafo issue for this related to [#14447](http://projects.theforeman.org/issues/14447). Also a mentioning in [README](https://github.com/theforeman/kafo/blob/master/README.md#parser-cache) about multiple cache files and example how to specify it would be appeciated, ideally also in [example file](https://github.com/theforeman/kafo/blob/master/config/kafo.yaml.example#L20).

lib/kafo/parser_cache_reader.rb
komidore64: `cache` is the parsed file, whereas i'm using `cache_paths[i]` to get the filepath to display incase of an error..

lib/kafo/parser_cache_reader.rb
komidore64: good idea! i hadn't thought of that.

lib/kafo/parser_cache_reader.rb
ares: oh I see

README.md
```diff
@@ -965,15 +965,21 @@ not persisted between runs.
 
 ## Parser cache
 
-A cache of parsed Puppet modules and manifests can be created to skip the use
+One or more caches of parsed Puppet modules and manifests can be created to skip the use
```
ekohl: A quick word about the merging strategy might be useful. Is it a deep merge of just the first module that's found? What has precedence, highest or lowest in the list?

ares:
Thanks @komidore64, merged as e5127756ef4e703627dd4095559ee1c9c9c9a84a, I updated the commit message so it will close the issue, not just link it (refs vs fixes).

README.md
ares: Ah sorry, github didn't refresh and I missed this comment, I'll send a separate PR with more info.

ares: @ekohl does [this](https://github.com/theforeman/kafo/pull/169) make sense? @komidore64 could you please take a look and verify I don't lie? :-)

komidore64:
@ares right, thank you. Yes, I should have changed it to "fixes." Thanks for noticing that! :)

README.md
komidore64: [commented](https://github.com/theforeman/kafo/pull/169).

PR #168, opened 2016-08-19
Pin json_pure for odler rubies

ares:


domcleal:
Merged as 6bf5b72 (typo fixed). Thanks @ares.

PR #169, opened 2016-08-23
Refs #16230 - more info in README about multicache

ares:


README.md
```diff
@@ -967,7 +967,9 @@ not persisted between runs.
 
 One or more caches of parsed Puppet modules and manifests can be created to skip the use
 of kafo_parsers at runtime. This is useful when kafo_parsers doesn't support the
-version of Puppet in use, and may also provide a small performance benefit.
+version of Puppet in use, and may also provide a small performance benefit. When multiple
+cache files are used, they are being loaded in order they are specified in the config
+file. If their files overlap, the later overrides earlier.
```
ekohl: What does override mean here? Fully replaces the older file or are they merged?

ares: [fully](https://github.com/theforeman/kafo/blob/e5127756ef4e703627dd4095559ee1c9c9c9a84a/lib/kafo/parser_cache_reader.rb#L29) I believe

komidore64: @ares yeah, that's a correct description of how it works. 

Here I have two fake extremely simple parser caches.

```yaml
# fake_foreman.yaml
---
  !ruby/sym version: 1
  !ruby/sym files:
    foo: foo
    bar: bar
    baz: baz
```

```yaml
# fake_katello.yaml
---
  !ruby/sym version: 1
  !ruby/sym files:
    foo: different foo
    fizz: fizz
    buzz: buzz
```

You can see here how they are merged. Note that they both share a `"foo"` key, and the second key from `fake_katello.yaml` is what's accepted into the final cache because it comes second in the list of cache files.

```
Kafo::ParserCacheReader.new_from_file(['/home/vagrant/parser-cache-scratch/fake_foreman.yaml','/home/vagrant/parser-cache-scratch/fake_katello.yaml'])
=> #<Kafo::ParserCacheReader:0x00000001bbde78
 @cache={:version=>1, :files=>{"foo"=>"different foo", "bar"=>"bar", "baz"=>"baz", "fizz"=>"fizz", "buzz"=>"buzz"}}>
```

domcleal: Can you fix the wording to state "fully"? e.g. `If their files overlap, the later cache fully replaces the previous cache for each file/manifest."

ares:
Thanks, fixed.

domcleal:
Merged, thanks @ares.

ekohl:
I don't think fully replaces is correct. In the example provided by @komidore64 I'd expect the bar and baz keys not to be present in the result on a full replacement. How about: If their files overlap, the later files are merged into the result. Any already present keys are overwritten.

domcleal:
The confusion may come from the word "files" which is perhaps being used for both the cache file and each manifest file within a cache file. The README shouldn't be concerned about the format of the cache itself ("keys") but the effect on the cached manifest file list.

The loading will fully replace the cache for each manifest (file) if it exists - to say that the files are merged might suggest that the two manifest caches may be merged together, when the first is actually overwritten.

I think what you're getting at is that the _cache files_ are merged, rather than the _manifest files_ (which was what I intended at https://github.com/theforeman/kafo/pull/169#discussion_r76006569).

ekohl:
> I think what you're getting at is that the cache files are merged, rather than the manifest files (which was what I intended at #169 (comment)).

You're absolutely right. That's the source of my confusion.

PR #170, opened 2016-08-24
fixes #16259 - require file containing PARSER_CACHE_VERSION

domcleal:
This can't be reproduced directly from a git checkout as kafo.gemspec requires `kafo/version`, so the constant gets loaded. Either a published gem installation or removing the require from the gemspec will show up the error.

### DIFF
--- a/lib/kafo/parser_cache_writer.rb
+++ b/lib/kafo/parser_cache_writer.rb
@@ -1,8 +1,10 @@
+require 'kafo/version'
+
 module Kafo
   class ParserCacheWriter
     def self.write(modules)
       {
-        :version => 1,
+        :version => PARSER_CACHE_VERSION,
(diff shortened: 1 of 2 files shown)


ares:
Thanks @domcleal, merging, I'll release a new version right away

lib/kafo/parser_cache_writer.rb
```diff
@@ -1,8 +1,10 @@
+require 'kafo/version'
+
 module Kafo
   class ParserCacheWriter
     def self.write(modules)
       {
-        :version => 1,
+        :version => PARSER_CACHE_VERSION,
```
ares: :+1: thank you

komidore64:
@domcleal thanks for catching this. like you mentioned, i didn't catch it from a git checkout since it worked happily. :+1:

PR #171, opened 2016-08-25
Fixes #16301 - add key-value store

mbacovsky:
The store is loaded from <CONFIG_DIR>/../store.d/ or from path specified by config's `store_dir` The content is
contained in a yaml files that are loaded in alphabetical order.
The store is read-only during the run-time.
The content can be extended by installing files in the store.d
(e.g. from installer plugins). The instance of the store is in
KafoConfigure.store and can be used form hooks.

Standlaone scripts need to create its own instance by providing the
path. The store class is lightweight and is not dependent on configured Kafo.


mbacovsky:
I pushed this to speed up the review, but I need to test it with katello installer and the scripts if it provides all the fetures we will need

ekohl:
Is this to maintain state between installer runs? Like some action has been performed?

mbacovsky:
@ekohl this is intended rather for sharing data (that are not intended to be configurable by user) among installer and installer related scripts. For now I'd like to use it for keeping the server name ('Katello', 'Foreman' or 'Satellite') and a few other strings to be displayed in the installer and scripts output.

ares:
is this still "DO NOT MERGE"?

mbacovsky:
@ares, now the store has all the features necessary to complete my other PRs

ares:
Thanks @mbacovsky, merging!

PR #172, opened 2016-08-26
Fixes #16342: Pass parser cache directly to reader

ehelms:


lib/kafo/configuration.rb
```diff
@@ -264,7 +264,7 @@ def migrations_dir
 
     def parser_cache
       if app[:parser_cache_path]
-        @parser_cache ||= Kafo::ParserCacheReader.new_from_file(File.expand_path(app[:parser_cache_path]))
+        @parser_cache ||= Kafo::ParserCacheReader.new_from_file(app[:parser_cache_path])
```
komidore64: Since we are removing `File.expand_path` here, we should probably add it into `#new_from_file` mapped to each path that ParserCacheReader is being handed to still account for relative paths.

ehelms: Added a line https://github.com/theforeman/kafo/pull/172/files#diff-3f80fc96ce636e776708977123694238R7  to expand each path.

komidore64: perfect.

A++ would comment again

ehelms: Had to update it based on the failures in unit tests -- https://github.com/theforeman/kafo/pull/172/files#diff-3f80fc96ce636e776708977123694238R18

ehelms:
@ares if you get a moment, this appears to block testing when the parser cache is specified as an array

domcleal:
Ticket number should be #16342, not 432.

ehelms:
@domcleal updated commit message

domcleal:
Thanks @ehelms.

PR #173, opened 2016-08-30
fixes #16381 - add module_enabled? helper to hook context

domcleal:


lib/kafo/configuration.rb
```diff
@@ -88,6 +88,10 @@ def modules
       @modules ||= @data.keys.map { |mod| PuppetModule.new(mod, PuppetModule.find_parser, self).parse }.sort
     end
 
+    def module(name)
+      modules.find { |m| m.name == name }
```
ares: could you please use the same method at https://github.com/domcleal/kafo/blob/e572fe236f19316f7ab97b12ac65d39831e3743d/lib/kafo/configuration.rb#L192

lib/kafo/hook_context.rb
```diff
@@ -61,6 +61,13 @@ def add_module(module_name, mapping = nil)
       self.kafo.add_module(module_name)
     end
 
+    # Check if a module is enabled in the current configuration.
+    # examples:
+    #   module_enabled?('example')
+    def module_enabled?(module_name)
+      self.kafo.module(module_name).enabled?
```
ares: I think you can use [this](https://github.com/domcleal/kafo/blob/e572fe236f19316f7ab97b12ac65d39831e3743d/lib/kafo/configuration.rb#L169) directly

ares:
thanks @domcleal, two small suggestions, test failure seems unrelated

lib/kafo/hook_context.rb
domcleal: `Configuration#module_enabled?` isn't authoritative, as the PuppetModule itself contains an enable/disable flag which takes precedence. When a module's disabled from the CLI or wizard, the flag on the PuppetModule is toggled, while the data hash (from the answers file) in `Configuration` remains.

I'm wary of changing how `Configuration#module_enabled?` works (e.g. to check the module status) as `PuppetModule.enabled?` already calls back to this method.

In short: the enabled? method on PuppetModule itself is the main entrypoint.

ares: oh, I forgot module can change the value of `@enabled`, from looking at the `enabled?` method it seems it's only memoization of  `Configuration#module_enabled?`. ok :+1: for this change then.

lib/kafo/configuration.rb
domcleal: Updated, now #param uses this method to find the module, then calls params on it.

domcleal:
PR updated to change Configuration#param to use the new #module method. This affected a few tests that stubbed #params instead of #modules, so that's fixed. Tests that stubbed both params/modules have been simplified to only stub modules.

test/kafo/configuration_test.rb
```diff
@@ -58,8 +58,8 @@ module Kafo
 
     describe '#params_changed' do
       it 'lists all the params that changed value' do
-        basic_config.stub(:params, [p_foo, p_bar, p_baz]) do
-          old_config.stub(:params, [p_old_foo, p_old_bar]) do
+        basic_config.stub(:modules, [fake_module('mod', [p_foo, p_bar, p_baz])]) do
```
ares: sorry I haven't realized this will need so many stub changes

ares:
Thanks @domcleal, merging.

PR #174, opened 2016-09-02
fixes #16006 - always define kafo_add_progress variable

domcleal:


ares:
Thanks @domcleal, fixes the warning for me, merging!

PR #175, opened 2016-09-02
fixes #6138 - raise error if --mod-param value is missing

domcleal:
When "--mod-param --mod-param2 foo" is given, the second parameter is
taken by clamp as the value of the first. Detect when a known option
switch is given as an argument.

### DIFF
--- a/lib/kafo/exit_handler.rb
+++ b/lib/kafo/exit_handler.rb
@@ -17,7 +17,8 @@ module Kafo
           :unknown_module => 24,
           :defaults_error => 25,
           :unset_scenario => 26,
-          :scenario_error => 27
+          :scenario_error => 27,
+          :missing_argument => 28
       }
     end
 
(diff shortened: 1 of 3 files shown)


lib/kafo/exit_handler.rb
```diff
@@ -17,7 +17,8 @@ def error_codes
           :unknown_module => 24,
           :defaults_error => 25,
           :unset_scenario => 26,
-          :scenario_error => 27
+          :scenario_error => 27,
+          :missing_argument => 28
```
ares: @iNecas any comment for alignment? :-)

ares:
Thanks @domcleal, merging!

ares:
Thanks @domcleal, merging (with small change in commit message)

lib/kafo/exit_handler.rb
iNecas: `:unset_argument` would just fit there :)

ares: sorry, too late, please open a separate PR :-P

PR #176, opened 2016-09-06
fixes #16445 - send multi-line Puppet messages to last seen log level

domcleal:


ares:
Thanks @domcleal, merging!

PR #177, opened 2016-09-14
fixes #16550 - load DataType where used in Param

domcleal:
~~There's still another issue found in our tests with 1.0.0, so don't worry about a release just yet.~~


komidore64:
ack

ares:
Merging, thanks @domcleal

PR #180, opened 2016-09-15
Fixes #16573 - don't run validations on disabled modules

iNecas:
I haven't added tests as there is no existing test
for KafoConfigure and required infrastructure and efford
for doing so would be outside of scope for this simple patch.


domcleal:
Thanks @iNecas.

PR #182, opened 2016-09-19
Fix tests for puppet 4

ares:


ares:
I'd appreciate if someone could take a look, this will allow us to enable puppet 4 on jenkins.

kafo.gemspec
```diff
@@ -23,11 +23,14 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'simplecov', '< 0.9'
   spec.add_development_dependency 'ci_reporter', '~> 1.9.0'
 
+  spec.add_development_dependency 'puppet'
+  # this is required if puppet is greater than 4.0
+  # spec.add_development_dependency 'puppet-strings'
```
domcleal: Should this be in the Gemfile so it's easier to enable? I assume you're suggesting devs would have to uncomment it.

test/kafo/hiera_configurer_test.rb
```diff
@@ -15,7 +15,7 @@ module Kafo
     end
 
     describe "#generate_data" do
-      let(:puppet_module) { PuppetModule.new('testing', TestParser.new(BASIC_MANIFEST)).tap { |m| m.enable }.parse }
+      let(:puppet_module) { @@puppet_module ||= PuppetModule.new('testing', TestParser.new(BASIC_MANIFEST)).tap { |m| m.enable }.parse }
```
domcleal: Should the caching live inside TestParser? Some of the test factories already have class-level caches.

kafo.gemspec
ares: if I add it there, jenkins is unhappy that we specify puppet twice (it adds it to Gemfile too). Maybe we could move the logic into Gemfile - for puppet >= 4 install puppet-strings, but I'm not sure if we can detect current puppet version in the middle of Gemfile, it might not be installed by bundler yet

ares:
[test]

ares:
http://ci.theforeman.org/job/test_kafo_master_pull_request/316/ succeeded, another run triggered by comment here seems to fail because of some network problem so it should be fine, the Gemfile now works the same way as for kafo_parsers, once we merge this I'll fix the master job in a same way

domcleal:
Merged, thanks @ares.

PR #183, opened 2016-10-07
fixes #16830 - replace HighLine::String inside arrays/hashes

domcleal:


ares:
Thanks @domcleal, merging!

PR #184, opened 2016-10-11
fixes #16862 - register custom data types from module path

domcleal:


ares:
it seems this does not work with puppet 4, not sure why but rdoc parsing fails on something

domcleal:
@ares if you mean the failing tests, that should be fixed by https://github.com/theforeman/kafo_parsers/pull/26 (puppet-strings 0.99 has a different schema to 0.4).

domcleal:
I've moved the registration of data types from KafoConfigure into Configuration#modules so they're registered when using kafo-export-params as well as the KafoConfigure installer runs. I was also able to add tests this way.

ares:
Thanks @domcleal, merging!

PR #187, opened 2016-10-26
fixes #17106 - change last scenario symlink to basename

domcleal:
Prevents relative paths from the CONFIG_DIR used by ScenarioManager
leaking into the last scenario symlink target, so it is now always a
relative target to a config file in the same directory. Broken symlinks
are handled gracefully now too.


ares:
Thanks @domcleal, merging.

PR #188, opened 2016-10-28
fixes #17090 - add acceptance tests for kafo-configure runs

domcleal:
Tests both general behaviour of kafo-configure and kafo-export-params by
generating a new installer project and using the local copy of Kafo. The
tests are not designed to be extensive or to replace unit tests, but to
offer basic reassurance that the installer functionality in
KafoConfigure works.

The speed is poor, mostly due to using Bundler for the installer and
Puppet, which means a development environment with additional
dependencies and fewer performance enhancements such as parser caches
and CFacter, so the number of tests should be kept to a minimum.

The first commit is from #187 to fix an issue found running these tests.


domcleal:
Setting to WoC as the minitest capture IO helper's throwing an error under ci_reporter (`rake jenkins:unit`).

ares:
also please rebase, #187 has been merged

domcleal:
PR rebased and the test issues fixed - the hashie gem needed pinning for Ruby 1.8.7 compatibility (it caused a subtle error inside powerbar), and I disabled stdout/err capturing for the jenkins:unit rake task as redefining $stdout/err conflicts with minitest's capture helpers. Stacktraces, assertion messages etc are all still captured correctly.

Tests are failing on Puppet 2.7.x because it doesn't have Hiera support. Since we already dropped support for 2.7 in e7b02e0, is it OK if I remove it from the Jenkins test matrices?

ares:
yup, please go ahead and update jenkins jobs

domcleal:
[test] - 2.7.x removed from both jobs.

ares:
Thanks @domcleal, merging. Very good addition. Since you're currently the most active contributor, would you like me to add you as a gem owner on rubygems? Don' worry, I'd keep doing releases :-)

domcleal:
Yes please, I'd appreciate that. (firstname [at] lastname.org is my rubygems e-mail address)

ares:
Welcome aboard!

PR #190, opened 2016-11-11
Pin json, rdoc gems on Ruby 1.x

domcleal:


mmoll:
merged, thanks @domcleal!

PR #191, opened 2016-11-11
fixes #4428 - automatically map namespaced classes to correct paths

domcleal:
Changes the default locations of namespaced classes (e.g. mymod::foo) to
match the Puppet autoloader locations, assuming the first component is
the module name and classes are laid out in the regular structure. Only
for non-standard layouts would the paths/names need to be overridden,
which modern Puppet versions have reduced support for in any case.

### DIFF
--- a/lib/kafo/puppet_module.rb
+++ b/lib/kafo/puppet_module.rb
@@ -138,12 +138,12 @@ module Kafo
 
     # custom module directory name
     def get_dir_name
-      mapping[identifier].nil? ? name : mapping[identifier][:dir_name]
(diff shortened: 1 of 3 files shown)


ares:
Works fine, thanks @domcleal, merging.

ares:
Would you mind preparing the foreman-installer patch that would simplify the mapping? I know this remains compatible but it would be great to use this with next version of kafo.

domcleal:
Yep, I have it - just waiting for the new version of Kafo to be available.

PR #192, opened 2016-11-11
fixes #3235 - replace ReportWrapper patch with Puppet's --evaltrace

domcleal:
Allows the progress bar to update before Puppet begins evaluating each
resource instead of after, and removes a monkey patch.


domcleal:
> The progress bar still has a rate-limited refresh cycle, so not every resource line will show up on the output.

I've fixed this by passing the force argument through to the bar.show() call from EVALTRACE_START and during resource prefetches, which should keep it fully up to date.

ares:
Nice, the logs are now more useful too. 3 years old feature request gets in, thanks @domcleal.

PR #193, opened 2016-11-24
fixes #17473 - parse empty string data type argument

domcleal:


ares:
Thanks also for good description in redmine issue, works fine, merging.

PR #194, opened 2016-11-30
fixes #6158 - only dump defaults for param defaults starting `$`

domcleal:
Refactors how defaults from the manifest and the defaults dump are
stored, keeping them in separate variables and expecting any params
with other variables as the default to have `$` prefixes.

Only params with a `$` default prefix will be dumped, so any other
default string values will not be dumped and are persisted correctly.

---

Needs https://github.com/theforeman/kafo_parsers/pull/31, I can increase the min kafo_parsers version in the gemspec if/when that's released.

### DIFF
--- a/lib/kafo/param.rb
+++ b/lib/kafo/param.rb
@@ -5,8 +5,10 @@ require 'kafo/validator'
 
 module Kafo
   class Param
-    attr_reader :name, :module, :type
-    attr_accessor :default, :doc, :value_set, :condition
+    UNSET_VALUES = ['UNSET', :undef]
+
+    attr_reader :name, :module, :type, :manifest_default
+    attr_accessor :doc, :value_set, :condition
     attr_writer :groups
 
     def initialize(builder, name, type)
@@ -20,8 +22,9 @@ module Kafo
     end
 
     # we use @value_set flag because even nil can be valid value
+    # Order of descending precedence: @value, @default (from dump), @manifest_default
     def value
-      @type.typecast(@value_set ? @value : default)
+      @value_set ? @type.typecast(@value) : default
     end
 
     def value=(value)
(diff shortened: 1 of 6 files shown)


lib/kafo/param.rb
```diff
@@ -34,8 +35,34 @@ def unset_value
...
     end
 
+    # For literal default values, only use 'manifest_default'. For variable values, use the value
+    # loaded back from the dump in 'default'.
+    def default
+      @type.typecast(dump_default_needed? ? @default : manifest_default)
+    end
+
+    def default=(default)
+      default = nil if ['UNSET', :undef].include?(default)
```
ares: could you extract `['UNSET', :undef]` from here and `manifest_default=` to some constant, e.g. UNSETTING_VALUES so they do not get out of sync?

lib/kafo/param.rb
```diff
@@ -34,8 +35,34 @@ def unset_value
...
+      @type.dump_default(manifest_default_params_variable)
+    end
+
+    def manifest_default=(default)
+      default = nil if ['UNSET', :undef].include?(default)
+      @manifest_default = default
+    end
+
+    def manifest_default_params_variable
+      manifest_default[1..-1] if manifest_default.to_s.start_with?('$')
```
ares: `if dump_default_needed?` or rename/alias the method so it makes sense in this context

lib/kafo/param.rb
domcleal: Extracted to `UNSET_VALUES`.

lib/kafo/param.rb
domcleal: Used `dump_default_needed?`.

domcleal:
PR updated with the suggested changes.

PR #196, opened 2016-12-01
fixes #15903 - show current vs default values in help output

domcleal:
Unifies display of default/value in interactive mode with help output.


ares:
Works very well, could you please rebase it so we get tests output for all versions? Otherwise RfM

domcleal:
PR rebased.

ares:
Thanks @domcleal, merging.

PR #197, opened 2016-12-01
fixes #17504, #15619 - use generated puppet.conf when running Puppet

domcleal:


lib/kafo/kafo_configure.rb
```diff
@@ -411,23 +412,26 @@ def run_installation
       hiera.write_configs
       self.class.exit_handler.register_cleanup_path(hiera.temp_dir)
 
+      puppetconf = PuppetConfigurer.new(
+        'color'        => false,
+        'evaltrace'    => !!@progress_bar,
+        'hiera_config' => hiera.config_path,
+        'noop'         => !!noop?,
+        'profile'      => !!profile?,
+        'show_diff'    => false,
```
ares: based on jenkins finding, trailing comma does not work in Ruby 1.8

lib/kafo/puppet_configurer.rb
```diff
@@ -0,0 +1,37 @@
...
+    end
+
+    def []=(key, value)
+      @settings[key] = value
+    end
+
+    def write_config
+      @logger.debug("Writing Puppet config file at #{@temp_file.path}")
+      @temp_file.open
+      @temp_file.truncate(0)
```
ares: just out of curiosity, is this because of race condition even when using `Tempfile`?

ares:
Except the 1.8 issue it looks good.

lib/kafo/puppet_configurer.rb
domcleal: It's in case the method is called twice so the file gets overwritten fully.

domcleal:
PR updated with the trailing comma removed.

ares:
Thanks @domcleal, merging.

PR #198, opened 2016-12-18
Fixes #16148 - Reword 'puppet module' to plugin

shlomizadok:
When reading the enable message, it can be misleading whether we enable
a plugin or a puppet module

### DIFF
--- a/lib/kafo/kafo_configure.rb
+++ b/lib/kafo/kafo_configure.rb
@@ -314,7 +314,7 @@ module Kafo
       modules.each do |mod|
         self.class.option d("--[no-]enable-#{mod.name}"),
                           :flag,
-                          "Enable '#{mod.name}' puppet module",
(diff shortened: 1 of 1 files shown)


domcleal:
These are Puppet modules, not plugins. Not sure this is any more correct either for Foreman or in a generic sense.

Can you use the text from the class documentation?

shlomizadok:
@domcleal - trying to keep it simple - so I have reworded to `Enable '#{mod.name}' module`.
Do you think it should be more descriptive?

ares:
It seems that the original issue had a different meaning. Let's postpone this until the reporter makes it clearer.

ares:
@shlomizadok thanks for the patch but it turns out another solution would fit better, I would prefer not merging this if you don't mind

ares:
I'm closing this, please reopen if you think the patch is still valid.

PR #200, opened 2017-01-04
fixes #14946 - add --[no-]parser-cache argument to force cache

domcleal:
With --parser-cache, outdated cache entries will be forced if present,
even if a parser is available. With --no-parser-cache, cache entries
will be ignored even if they appear up to date (same as the existing
kafo-export-params argument).

The default behaviour with neither argument remains the same, to only
use the cache when it appears current.

### DIFF
--- a/lib/kafo/parser_cache_reader.rb
+++ b/lib/kafo/parser_cache_reader.rb
@@ -2,6 +2,8 @@ require 'kafo/version'
 
 module Kafo
   class ParserCacheReader
+    attr_accessor :force
+
     def self.new_from_file(cache_paths)
       cache_paths = [cache_paths].compact unless cache_paths.is_a?(Array)
       if cache_paths.empty?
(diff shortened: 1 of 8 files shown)


lib/kafo/parser_cache_reader.rb
```diff
@@ -38,20 +40,30 @@ def self.logger
...
+        return nil
+      end
+
       return nil unless @cache[:files].has_key?(key)
 
       if @cache[:files][key][:mtime] && File.mtime(manifest_path).to_i > @cache[:files][key][:mtime]
-        logger.debug "Parser cache for #{manifest_path} is outdated, ignoring cache entry"
-        return nil
+        if @force
+          logger.debug "Parser cache for #{manifest_path} is outdated, forced to use it anyway"
```
ares: if this should be used with caution, would it make more sense to use warning instead of debug log level here?

domcleal: Yes, that's a good point. PR updated with warn-level logging on this statement.

ares:
Thanks @domcleal, merging.

PR #201, opened 2017-01-05
Add Kafo execution schema to the docs

mbacovsky:
As a part of some investigation I created diagram of Kafo execution with the important points and events marked.


ares:
Thanks @mbacovsky, merging.

PR #202, opened 2017-01-06
fixes #17967 - re-enable show_diff setting

domcleal:


ares:
Thanks @domcleal, merging.

PR #203, opened 2017-01-06
fixes #17956 - load missing defaults from data in Puppet modules

domcleal:
For parameters without a default value and under Puppet 4, try calling
the lookup() function for each parameter name to retrieve any values
defined in the module data (e.g. Hiera files under data/). This allows
module authors to use the modern version of the 'params' pattern to
define complex, conditional default values.

New Puppet functions and a wrapper class merges the dumped values from
lookup() with the dump of params values, and output it via notice().

Demonstration: https://www.youtube.com/watch?v=FqAzIL8ENFg

### DIFF
--- a/README.md
+++ b/README.md
@@ -552,6 +552,40 @@ as key:value.
 When parsing the value, the first colon divides key and value. All other
 colons are ignored.
 
+## Default values
+
+Default values for parameters are read from the class definitions in the
+manifests. If values are given inline then these will be stored by Kafo as the
+initial value of the parameter (unless set in the answers file or later changed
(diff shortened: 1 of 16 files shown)


domcleal:
PR updated to fix the test failures. Due to PUP-6230 causing undef data values to return `{}` which is used in the tests, I restricted runtime support to Puppet 4.5.0 and above.

ares:
Thanks @domcleal, I tested on existing installers with older puppet and it works well. Also the remaining ":default_values_dir:" in config file does not do any harm. Merging!

PR #207, opened 2017-03-15
Fixes #18910 - pin clamp to 1.8 compatible version

ares:


kafo.gemspec
```diff
@@ -30,7 +30,7 @@ Gem::Specification.new do |spec|
   # better logging
   spec.add_dependency 'logging', '< 3.0.0'
   # CLI interface
-  spec.add_dependency 'clamp', '>= 0.6.2'
+  spec.add_dependency 'clamp', '>= 0.6.2', '< 1.1.2'
```
domcleal: This should be in `Gemfile`, like other version-dependent pins.

ares:
Thanks @domcleal, moved there and update the commit message. Please take another look, the job seems to be have some green runs on 1.8 already http://ci.theforeman.org/job/test_kafo_master_pull_request/370/

domcleal:
Merged, thanks @ares.

PR #209, opened 2017-03-16
Refs PR 201 - fix files included

ares:
Fixes the problem from #201. Originally I was thinking to not ship the png but since it's linked in readme and it's only 160k I think that's fine.


domcleal:
Merged, thanks @ares.

PR #210, opened 2017-03-16
refs #17956 - compare version properly for Puppet 4.10.x compat

domcleal:


test/acceptance/kafo_configure_test.rb
```diff
@@ -131,7 +130,7 @@ module Kafo
       end
 
       it 'must create file' do
-        skip 'Requires Puppet 4.5+ for data in modules' if Puppet::PUPPETVERSION < '4.5.0'
+        skip 'Requires Puppet 4.5+ for data in modules' if Gem::Specification.find_all_by_name('puppet').sort_by(&:version).last.version < Gem::Version.new('4.5.0')
```
ares: why searching for all puppet versions? is Puppet::PUPPETVERSION removed in 4.10? I'd expect we'd need to do something like `Gem::Version.new(Puppet::PUPPETVERSION)`

domcleal: I used the same comparison method as other tests just for consistency, which is to check the gem version rather than require the library and check its version. If you wish, I can refactor them all to use `puppet/version`.

ares: Ah ok, makes sense. I don't recall if I had a reason to write it like that in other tests. Refactoring would be nice but not required in this PR. I'll merge and do the release, thank you.

ares:
Thanks @domcleal , merged.

PR #212, opened 2017-04-10
fixes #19215 - parse nested data type arguments accurately

domcleal:
No longer greedily matches data type arguments until the last closing
square bracket, ensuring arguments are split correctly. Changed to a
string scanning implementation so it's able to parse nested brackets
without relying on the greedy regexp (or recursion).

### DIFF
--- a/lib/kafo/data_type.rb
+++ b/lib/kafo/data_type.rb
@@ -1,3 +1,5 @@
+require 'strscan'
+
 module Kafo
   class DataType
     def self.new_from_string(str)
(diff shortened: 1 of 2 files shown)


lib/kafo/data_type.rb
```diff
@@ -39,16 +41,42 @@ def self.unregister_type(keyword)
...
+        else # bare words, or Type::Name, or Type::Name[args..]
+          type = scanner.scan(/[\w:-]+/) or raise ConfigurationException, "missing argument #{args.count + 1} to data type #{input}"#
+
+          # store inner arguments as a continuation of the type string
+          if scanner.peek(1) == '['
+            type << scanner.getch
+            bracket_count = 1
+            until bracket_count.zero?
+              next_bracket = scanner.scan_until(/[\[\]]/) or raise ConfigurationException, "missing close bracket in argument #{args.count + 1} in data type #{input}"
+              case next_bracket[-1..-1]
```
ares: isn't this the same as `next_bracket[-1]` or `scanner.peek(1)`? or is that because older ruby would return character ord value?

domcleal: Yes, it's because of Ruby 1.8.7 returning the ord value - this makes it easy to compare the character value in the case statement.

ares:
Thanks @domcleal, merging.

PR #213, opened 2017-04-12
fixes #19254 - check Puppet version is supported by modules

domcleal:
Using the Puppet version requirements from metadata.json in the
available modules, the running Puppet version is validated to ensure it
is compatible. The user can skip this with `--skip-puppet-version-check`
if they wish.

### DIFF
--- a/lib/kafo/puppet_command.rb
+++ b/lib/kafo/puppet_command.rb
@@ -6,22 +6,16 @@ module Kafo
       @command = command
       @puppet_config = puppet_config
 
-      # Expand the modules_path to work around the fact that Puppet doesn't
-      # allow modulepath to contain relative (i.e ..) directory references as
-      # of 2.7.23.
-      @options = options.push("--modulepath #{File.expand_path(modules_path)}")
+      @options = options.push("--modulepath #{modules_path.join(':')}")
       @options.push("--config=#{puppet_config.config_path}") if puppet_config
       @logger  = KafoConfigure.logger
-    end
-
-    def add_progress
-      %{$kafo_add_progress="#{!KafoConfigure.verbose}"}
+      @puppet_version_check = !configuration.app[:skip_puppet_version_check]
     end
 
     def command
       @puppet_config.write_config if @puppet_config
       result = [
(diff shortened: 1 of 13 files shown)


lib/kafo/puppet_command.rb
```diff
@@ -44,11 +38,65 @@ def self.search_puppet_path(bin_name)
...
+
+    def generate_version_checks
+      checks = []
+      modules_path.each do |modulepath|
+        Dir[File.join(modulepath, '*', 'metadata.json')].sort.each do |metadata_json|
+          metadata = JSON.load(File.read(metadata_json))
+          next unless metadata['requirements'] && metadata['requirements'].is_a?(Array)
+
+          metadata['requirements'].select { |req| req['name'] == 'puppet' }.each do |req|
+            checks << versioncmp(metadata['name'], req['version_requirement'])
```
ekohl: Should this block als check if name and version_requirement are present or should it assume the metadata is valid?

domcleal: Thanks, I've added a check for `version_requirement` into the `select` on the previous line, so this block will only run if both are set.

ares:
The PR looks good and works fine. This causes problems with foreman-installer running on puppet 3.8.6 and puppet-extlib 1.1.0 which requires 3.8.7+. I know there's a big group of users with this combination so I want to first make sure there's some better solution than asking them to run the installer with `--skip-puppet-version-check`. If you don't mind I'd do the release with what we have so far and keep this unmerged before it's solved.

domcleal:
Thanks, I wasn't aware of that incompatibility. Foreman installer 1.16.x will not support Puppet 3 at all, so it shouldn't matter that much - it would only be a problem with the modules shipped in 1.15 or older.

If something similar happened in the future, we would now be a little more aware of it through testing and could either correct the module metadata, or ship a version of the module that does claim to be compatible. Otherwise yes, a user would have to run with the skip argument, but I wouldn't expect that to be the norm.

ares:
Makes sense, I wanted to do a fix release compatible with 1.15

I think the best will be to bump the version to 2.1.0 once this is merged so if someone will require more fixes to be used with 1.15 we can always continue with 2.0 branch.

ares:
Thanks @domcleal, merging.

PR #214, opened 2017-04-21
fixes #19341 - remove invalid UTF-8 encoded chars from Puppet log

domcleal:


fatmcgav:
@domcleal That was quick :D 

Happy to test if can give me some pointers :)

domcleal:
Applying the patch to your file at /usr/share/gems/gems/kafo-1.0.5/lib/kafo/puppet_log_parser.rb should fix it. I tested with the fact value in the Redmine issue and it did appear to run correctly again.

fatmcgav:
Of course :) 

Applied the patch, and can confirm that foreman-installer ran through successfully. 

Cheers again for the quick fix...

ares:
Thanks @domcleal, merging.

PR #218, opened 2017-09-18
Fix exit() in hook_context and a deprecation warning

ekohl:
I didn't know how to write a proper test for this so I'm submitting this as is now. Pointers on how to properly write them would be appreciated.

### DIFF
--- a/lib/kafo/exit_handler.rb
+++ b/lib/kafo/exit_handler.rb
@@ -34,7 +34,7 @@ module Kafo
     end
 
     def translate_exit_code(code)
-      return code if code.is_a?(Fixnum)
+      return code if code.is_a?(Integer)
       if error_codes.has_key?(code)
         return error_codes[code]
       else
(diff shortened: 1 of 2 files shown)


lib/kafo/exit_handler.rb
```diff
@@ -34,7 +34,7 @@ def exit(code, &block)
     end
 
     def translate_exit_code(code)
-      return code if code.is_a?(Fixnum)
+      return code if code.is_a?(Integer) || code.is_a?(Fixnum)
```
mmoll: will there be cases, where Integer doesn't suffice?

ekohl: Good question. I'm not that familiar with Ruby so I figured this was safer but it's supposed to be a base class so is just Integer sufficient?

mmoll: I would say yes.

ares:
Thanks @ekohl, works well. `Integer` is the right class IMHO. Merging.

PR #219, opened 2017-09-18
Expand hook contexts

ekohl:
This moves facts to a BaseContext so it can be used both in migrations and in hooks. In addition to that the scenario is provided in the hook context.

The facter output is now JSON rather than YAML.

With this the facts are now recursively processed so we can support structered facts with symbols as well. This might be a breaking change if users are already using structured facts.

### DIFF
--- a/lib/kafo/base_context.rb
+++ b/lib/kafo/base_context.rb
@@ -1,0 +1,35 @@
+module Kafo
+  class BaseContext
+    def facts
+      self.class.facts
+    end
+
+    private
+
+    def self.symbolize(data)
+      case data
+      when Hash
+        Hash[data.map { |key, value| [key.to_sym, symbolize(value)] }]
+      when Array
+        data.map { |v| symbolize(v) }
+      else
+        data
+      end
+    end
+
+    def self.clear_caches
+      @facts = nil
+      @facter_path = nil
+    end
+
+    def self.facts
+      @facts ||= begin
+        symbolize(JSON.load(`#{facter_path} --json`) || {})
+      end
+    end
+
+    def self.facter_path
+      @facter_path ||= PuppetCommand.search_puppet_path('facter')
+    end
+  end
+end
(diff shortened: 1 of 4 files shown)


lib/kafo/base_context.rb
```diff
@@ -0,0 +1,37 @@
...
+    def facts
+      self.class.facts
+    end
+
+    private
+
+    def self.symbolize(data)
+      case data
+      when Hash
+        {}.tap do |h|
```
ares: I don't insist but this 3 lines could be written as

```ruby
Hash[data.map { |key, value| [key.to_sym, symbolize(value)] }]

Comment (ekohl) [reply lib/kafo/base_context.rb]: My lack of Ruby experience is showing. Ty, that's a pattern I wanted to have but didn't know how :)

**Comment by ekohl** [comment]
> Nice work, why is json prefered over yaml format?

Ohad suggested it would be faster but I have no strong opinion on that.

File: lib/kafo/hook_context.rb [hunk 2]
```diff
@@ -90,5 +91,10 @@ def get_custom_config(key)
     def store_custom_config(key, value)
       self.kafo.config.set_custom(key, value)
     end
+
+    # Return the path to the current scenario
+    def scenario
```
ekohl: Should his be ```scenario_path``` and should we add a parsed version as well?

ekohl:
[test]

lib/kafo/hook_context.rb
ares: sounds good to me, since this is sort of an API for installers, we should make it right on first attempt :-)

ares:
Thanks @ekohl, I ran tests locally and they pass just fine. Merging so we don't have to wait until we resolve jenkins problems.

ekohl:
@ares I didn't change the scenario to scenario_path yet. Should I submit a new PR for that to fix it before a release?

ares:
Yes please, I though that would be a separate method anyway.

ekohl:
https://github.com/theforeman/kafo/pull/224 should be a cleaner API

PR #221, opened 2017-09-19
Refs #20893 - Rename file after function got renamed

mmoll:


ares:
Thanks @mmoll, merging.

PR #223, opened 2017-09-22
Fixes #21078 - Ensure all hook dirs are included

ehelms:


ares:
Thanks @ehelms, I ran tests locally and all seems fine, merging so we don't have to wait until we resolve jenkins issues.

PR #224, opened 2017-09-25
Fixes #21064 - Fix scenario context in hooks

ekohl:


test/kafo/hook_context_test.rb
```diff
@@ -11,6 +11,24 @@ module Kafo
...
+      specify { context.respond_to?(:get_custom_config).must_equal true }
+      specify { context.respond_to?(:store_custom_config).must_equal true }
+      specify { context.respond_to?(:scenario_path).must_equal true }
+      specify { context.respond_to?(:scenario_data).must_equal true }
+    end
+
+    describe "#scenario_data" do
+      specify do
+        Tempfile.open('scenario') do |file|
+          file.write(YAML.dump({foo: 'bar'}))
```
ares: this wouldn't work with Ruby 1.8 which is still used with puppet 3.8, would you mind using hash rocket `=>` syntax here please?

test/kafo/hook_context_test.rb
```diff
@@ -11,6 +11,24 @@ module Kafo
...
+          file.flush
+
+          context.stub :scenario_path, file.path do
+            assert_equal context.scenario_data, {foo: 'bar'}
```
ares: same as above

test/kafo/hook_context_test.rb
ekohl: Updated

ares:
Thanks @ekohl, merging after running tests locally.

PR #226, opened 2018-05-11
Allow skipping specific checks

ekohl:
There is still an odd interaction where every skipped check is duplicated in `skip_check_list`. Haven't tracked it down yet.


ares:
Seems tests need to be updated

ares:
[test kafo]

ekohl:
https://projects.theforeman.org/issues/29459

PR #227, opened 2018-09-11
Do not use global environmentpath

mattock:
This prevents the Kafo installer from crashing if the default (global)
environment, typically /etc/puppetlabs/code/environments/production, includes a
classes array that contains classes not provided by the installer. This is often
the case when running the Kafo installer on a production Puppetmaster where
environments have been deployed.

### DIFF
--- a/lib/kafo/kafo_configure.rb
+++ b/lib/kafo/kafo_configure.rb
@@ -429,12 +429,13 @@ module Kafo
       self.class.exit_handler.register_cleanup_path(hiera.temp_dir)
 
       puppetconf = PuppetConfigurer.new(
-        'color'        => false,
-        'evaltrace'    => !!@progress_bar,
-        'hiera_config' => hiera.config_path,
-        'noop'         => !!noop?,
-        'profile'      => !!profile?,
-        'show_diff'    => true
+        'color'           => false,
+        'evaltrace'       => !!@progress_bar,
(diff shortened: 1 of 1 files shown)


mattock:
This issue has been discussed [in more detail](https://community.theforeman.org/t/custom-kafo-based-puppetmaster-installer-fails-if-puppetmaster-has-a-yaml-flie-in-production-environment/10496/5) on Foreman Support forum. The resulting puppet.conf file looks like this:

```
[main]
color = false
environmentpath = /tmp/kafo_hiera20180911-30532-ekc6gy
evaltrace = true
hiera_config = /tmp/kafo_hiera20180911-30532-ekc6gy/hiera.conf
noop = false
profile = false
reports =
show_diff = true
```
This change has only been tested on Puppet 5.5.6. However, the environmentpath parameter seems to have been available as early as Puppet 3.x.

ares:
Thanks for the patch, @ekohl do you think this is going to work with foreman-installer in case hiera overrides are used?

ekohl:
I'm a bit unfamiliar with this part of kafo but I don't know if we really have environments. I'm wondering if it's good to set environmentpath to empty and only set the modulepath.

ekohl:
This is still on my radar, but haven't had time to verify it yet. It does look like we should also set `basemodulepath` to something we control to avoid things leaking into our environment.

PR #228, opened 2018-09-27
Fixes #25058 - More robust module_enabled?

ekohl:
If you have modules that are not available in every scenario. In the katello case there's katello_devel for example. Currently there's a workaround in the installer for when kafo.module returns nil but the hook itself can handle this as well. Unknown modules can be considered disabled.


ares:
Thanks @ekohl, merging

PR #229, opened 2018-11-19
Modernize kafo

ekohl:
* Require Ruby 2.0+
* Require Puppet 4.5+
* Remove support for pre-Puppet 4 data types
* Remove password support
* Fix warnings


ares:
I didn't find anything wrong with the code. While I see you made some subjective changes (e.g. exists? -> exist? or must_be_nil -> assert_nil), I'm fine with that as long as you will be the main maintainer :-) The test failures though seems valid (though failing randomly), do you happen to know what might be the cause?

Otherwise :+1:, could you also bump the major version?

ekohl:
The subjective changes were warnings I saw while running tests. Not sure where exactly they're from because I didn't see an explicit rubocop run.

And I couldn't quite pin down the failures. It looks like it's disabling the puppet module and I don't know why.

ekohl:
[test kafo]

ekohl:
[test kafo]

ares:
It seems as most of combinations are green now, probably good enough for considering this green?

mmoll:
[test kafo]

mmoll:
Hrm, still the "tries to connect to unreachable IP" problem, probably stemming from some fact, leading to timeouts.

ares:
When we checked that last time, it was contacting some ec2 host. I think that was some fact or something.

ekohl:
[test kafo]

ekohl:
It looks like there's some randomness in the test order and some tests depend on setups from other tests. I already fixed one in the last commit, but I'll need to trace down the rest.

ekohl:
[test kafo] the 2.0.0 build timed out after an hour. Looks like it's often on a single slave so if it repeats, I'll investigate it further. Other than that I think the build should pass now.

ekohl:
[test kafo] to see if dropping traffic using `iptables -A OUTPUT -d 169.254.169.254 -j REJECT` helps

ekohl:
:green_heart:

ekohl:
A real test is a good idea and I don't mind holding off merging until I have time for it.

ehelms:
Tested a Katello upgrade RPM install with this and all was green.

ares:
Thanks @ekohl, merging!

PR #230, opened 2018-12-10
[WIP] Fixes #25657 - Clear data types during configuration

ekohl:
It turns out that all data types are in a global registry. When Kafo::Configure is loaded initialized, it reads the data types. This is done when the scenario has changed. That's correct, but it should clear the global registry to avoid stale entries and potential duplicates.

### DIFF
--- a/lib/kafo/configuration.rb
+++ b/lib/kafo/configuration.rb
@@ -3,6 +3,7 @@ require 'yaml'
 require 'tmpdir'
 require 'kafo/puppet_module'
 require 'kafo/color_scheme'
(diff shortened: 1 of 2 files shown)


ekohl:
So fundamentally this doesn't work in its current form since it also clears the built in types.

ares:
I suppose most commits got already in, this needs to be rebased. Is it still WIP?

ekohl:
Rebased so the diff is clear, but this is still a fundamental issue. Kafo has built in data types that register themselves inside their class. If we clear this, we have no way to re-register them.

PR #231, opened 2019-02-20
Refactor kafo_configure module

ekohl:
This modernizes the kafo_configure module and allows it to work with Puppet 6. It still generates one deprecation warning about hiera_include which I'll follow up on.


ares:
Seems like some tests are failing

mmoll:
[test kafo]

ekohl:
It fails because I don't do the version checks before dumping the defaults. The defaults don't include hiera. That does show an interesting bug that I think some user ran into. If you use `--reset-PARAM` then you get the puppet default. If we used a different default somewhere in our hiera config, then you don't get that value. IMHO that's a bug rather than a feature

ekohl:
I now realize we never change values for parameters via hiera so we haven't ran into this yet. We always change it via answers

ekohl:
https://github.com/theforeman/kafo/pull/232 replaces most of this. I'd still like to pass in more data via hiera rather than generating puppet code on the command line, but that shouldn't block Puppet 6 support. I may revisit this at some point.

PR #237, opened 2019-03-12
Remove password manager

ekohl:
541f90af875bb2640c2f9ec2739182a3f56077fd removed password support, but failed to remove the PasswordManager.

IMHO this can be a 3.1.0 release since it's not really an API we intend to expose but I'm open to other arguments.


ares:
3.1.0 totally makes sense, it's not public API. We probably need to drop some tests as well.

ares:
[test kafo] looks like random failures

mmoll:
:green_heart:

ares:
Thanks @ekohl, merging!

PR #238, opened 2019-08-30
Add internal Hiera variables

ekohl:
Internal Hiera variables allow setting additional variables. This allows setting variables for unexposed modules.

Initially I had hoped to be able to use this in hiera.yaml but you can't use lookup functions there. Probably because that would introduce recursion. Hooks can still set a collection of settings but it'd still be nicer to keep them in separate files so they can easily be updated.

This can replace the hacky setup in the mongodb migration (https://github.com/theforeman/foreman-installer/blob/4160db8eb1d8b4eb58d30fbb4cb8477db990a3e0/katello/hooks/pre/31-mongo_storage_engine.rb#L58-L61 and https://github.com/theforeman/foreman-installer/blob/4160db8eb1d8b4eb58d30fbb4cb8477db990a3e0/katello/hooks/pre/30-upgrade.rb#L60-L64).

### DIFF
--- a/README.md
+++ b/README.md
@@ -783,6 +783,8 @@ Note that the hook is evaluated in HookContext object which provides a DSL:
 * ```get_custom_config``` and ```store_custom_config``` access custom config storage which persists
   among kafo runs
 * ```logger``` is also available for writing log messages
+* ```set_internal_hiera_variable```, ```get_internal_hiera_variable```,
+  ```unset_internal_hiera_variable``` can be used to set additional hiera variables.
 
 For more details, see
 [hook_context.rb](https://github.com/theforeman/kafo/blob/master/lib/kafo/hook_context.rb).
--- a/lib/kafo/configuration.rb
+++ b/lib/kafo/configuration.rb
@@ -29,6 +29,7 @@ module Kafo
         :custom               => {},
         :low_priority_modules => [],
         :verbose_log_level    => 'info',
+        :internal_hiera_variables => {},
         :skip_puppet_version_check => false
     }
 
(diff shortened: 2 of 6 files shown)


ekohl:
Looks like master is also failing tests. Will look into them next week.

ehelms:
A comment providing an example of howthis gets used by an installer would be great.

ekohl:
My WIP hooks that I tried to add to foreman-installer
```ruby
# katello/hooks/boot/13-hiera.rb
tuning_sizes = ['small', 'medium', 'large']
tuning_variable = 'kafo_configure::internal::tuning'

app_option(
  '--tuning',
  'INSTALLATION_SIZE',
  "Tune for an installation size. Choices: #{tuning_sizes.join(', ')}",
  :default => get_internal_hiera_variable(tuning_variable)
) do |value|
  # TODO: Validation doesn't appear to happen
  unless value.nil? || tuning_sizes.include?(value)
    signal_usage_error "only #{tuning_sizes.join(', ')} are valid tuning profiles"
  end
  value
end
```

```ruby
# katello/hooks/pre_commit/13-hiera.rb
tuning_variable = 'kafo_configure::internal::tuning'

current = get_internal_hiera_variable(tuning_variable)
new = app_value(:tuning)
if current != new
  set_internal_hiera_variable('kafo_configure::internal::tuning', new)
  kafo.config.configure_application
end
```

Then I found out I couldn't actually use `lookup()` in `hiera.yaml` so next week I'll retry that part by providing facts which I know for a fact can be used in `hiera.yaml`.

https://github.com/theforeman/foreman-installer/pull/383 would be an example of this mechanism.

PR #239, opened 2019-09-02
Pin clamp to < 1.3.1

ekohl:
Clamp 1.3.1 changed the help output method to first buffering it in an array of lines and then output it. It also used to override the puts method, but no longer does.


mmoll:
merged, bedankt @ekohl!

PR #240, opened 2019-09-02
Compatibility with clamp 1.3.1

ekohl:
Not sure this is the best approach. Currently Hammer requires clamp < 1.2.0 so in packaging this would break.

### DIFF
--- a/kafo.gemspec
+++ b/kafo.gemspec
@@ -32,7 +32,7 @@ Gem::Specification.new do |spec|
   # better logging
   spec.add_dependency 'logging', '< 3.0.0'
   # CLI interface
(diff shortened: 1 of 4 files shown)


ehelms:
This would mean carrying two versions of clamp on the RPM side? One for SCL and one for non-SCL? Have we checked with hammer team about updating clamp dependency?

ekohl:
I'm thinking about submitting a PR to clamp to unbreak the API. Essentially they made a breaking change in a patch release by removing the class method `puts`.

ekohl:
[test kafo] because I want to check the Jenkins job

PR #241, opened 2019-09-05
Introduce an ExecutionEnvironment + change Hiera loading + facts

ekohl:
This is a rather large PR. Individual commits have descriptions why they're useful. On a high level this shouldn't change Kafo except when using a Hiera config file. Then there is a change needed.

This will be Kafo 4.0 though I may add a few more things.

I'm going to test this but it'd be nice to get some initial review.


ehelms:
I interpret this as a re-factor with a bit of functionality added. The unclear part to me is what functionality has been added from the users perspective. In other words, how do I use the new functionality?

ekohl:
I don't know why the test is failing. I suspect it's wrapping somewhere but I don't know why. Looking at the matrix it's fairly random and that part of the code shouldn't be changed.

I'll also have a look at writing docs once I've verified the functionality works well rather than the basic testing I did now.

README.md
```diff
@@ -1004,6 +1004,18 @@ The cache will be skipped if the file modification time of the manifest is
 greater than the mtime recorded in the cache. Using `--parser-cache` will force
 the use of an outdated cache, but this should be used with caution.
 
+## Facts
+
+Kafo provides a structured fact describing the state. This fact is only present
+during the Puppet run. It is a structured fact containing the scenario name.
```
evgeni: You used `structured fact` twice, probably needs some rewording :)

README.md
```diff
@@ -1018,22 +1030,20 @@ The contents of this file are as per the
...
-missing, it will be added.
+The hierarchy must contain the path `%{facts.kafo.scenario.answer_file}` which
+contains all answers in a temporary location.
 
 As an example, a hierarchy could be set up with:
 
 ```yaml
 hierarchy:
   - name: "Kafo Answers"
+    path: "%{facts.kafo.scenario.answer_file}"
```
evgeni: That entry is not part of the above example?

README.md
```diff
@@ -1058,6 +1069,17 @@ hierarchy:
...
 
+The scenario name is also available as a fact which can be used to provide
+scenario specific overrides for unmanaged modules.
+
+```yaml
+hierarchy:
+  - name: "Kafo Answers"
+    path: "%{facts.kafo.scenario.answer_file}"
+  - name: "Scenario defaults"
+    path: "scenario/%{facts.kafo.scenario.name}.yaml"
```
evgeni: `kafo.scenario.name` in the above example is `Foreman`, but I'd expect to see `foreman.yaml` (lower-case F) for the filename?

ekohl: So right now `name` is the human readable name. I'm thinking about introducing a more usable one like `id` which would match the one passed on the command line.

lib/kafo/puppet_configurer.rb
```diff
@@ -23,15 +19,13 @@ def []=(key, value)
...
 
     def write_config
-      @logger.debug("Writing Puppet config file at #{@temp_file.path}")
-      @temp_file.open
-      @temp_file.truncate(0)
-      @temp_file.puts '[main]'
-      @settings.keys.sort.each do |key|
-        @temp_file.puts "#{key} = #{@settings[key]}"
+      @logger.debug("Writing Puppet config file at #{config_path}")
+      File.open(config_path, 'w') do |f|
```
ehelms: I'm less of a fan on multiline blocks of using a single character variable as it hurts readability of the code when something like `file` provides more context.

lib/kafo/configuration.rb
```diff
@@ -90,6 +91,14 @@ def set_custom(key, value)
       custom_storage[key.to_sym] = value
     end
 
+    def get_custom_fact(key)
```
ares: in ruby we typically do this

```
def custom_fact(key)
  custom_fact_storage[key.to_s]
end

def custom_fact=(key, value)
  custom_fact_storage[key.to_s] = value
end
```

but given method names above whose author probably knew what he was doing (:man_facepalming:) I'm happy to keep as is, just wanted to let you know in case you haven't seen this before

lib/kafo/execution_environment.rb
```diff
@@ -0,0 +1,78 @@
...
+      @directory ||= begin
+        directory = Dir.mktmpdir('kafo_installation')
+        @logger.debug("Creating execution environment in #{directory}")
+        directory
+      end
+    end
+
+    def store_answers
+      answer_data = HieraConfigurer.generate_data(@config.modules, @config.app[:order])
+      File.open(answer_file, 'w') { |f| f.write(YAML.dump(answer_data)) }
```
ares: how about debug log with information "writing to file #{answer_file}" on line above this one?

lib/kafo/exit_handler.rb
```diff
@@ -1,11 +1,10 @@
 module Kafo
   class ExitHandler
-    attr_accessor :cleanup_paths, :exit_code, :logger
```
ares: this drops `logger=`, which could still be easily added by `attr_writer :logger`, but if we don't use it anywhere (e.g. tests) I have no problem with it

lib/kafo/fact_writer.rb
```diff
@@ -0,0 +1,24 @@
+module Kafo
+  class FactWriter
+    DATA_FILENAME = 'kafo.yaml'
+    WRAPPER_FILENAME = 'kafo.rb'
+
+    def self.write_facts(facts, directory)
+      Dir.mkdir(directory)
+
+      # Write a data file containing all the facts encoded as YAML
+      File.open(File.join(directory, DATA_FILENAME), 'w') { |f| f.write(YAML.dump(facts)) }
```
ares: would be nice to have debug log for this (similar to elsewhere)

same few lines below

lib/kafo/hiera_configurer.rb
```diff
@@ -1,111 +1,37 @@
...
-      end
-
-      logger.debug("Creating Hiera data files in #{data_dir}")
-      FileUtils.mkdir(data_dir)
-
-      File.open(File.join(data_dir, HIERARCHY_FILENAME), 'w') do |f|
-        f.write(format_yaml_symbols(generate_data(@modules).to_yaml))
-      end
+    def self.write_default_config(path)
+      File.open(path, 'w') { |f| f.write(YAML.dump(default_config)) }
```
ares: log with file path would be great

lib/kafo/configuration.rb
ekohl: I've seen it but decided to copy what was already there for a consistent API.

lib/kafo/exit_handler.rb
ekohl: I think the logger should not be part of the public API. Tests create a new exit handler every time so it shouldn't be that relevant.

lib/kafo/fact_writer.rb
ekohl: Rather than adding logging to the FactWriter, I've done this in the ExecutionEnvironment.

lib/kafo/hiera_configurer.rb
ekohl: Also added the log line in ExecutionEnvironment.

ekohl:
I believe this is now ready for review. I'm going to update the foreman installer to use the new version. Ideally this should be released this week so it can start to soak in nightly.

lib/kafo/exit_handler.rb
ares: Ack

ares:
I have no more comments, if there's no concern raised over night and tests are green, I'll merge.

mmoll:
[test kafo]

test/kafo/execution_environment_test.rb
```diff
@@ -0,0 +1,43 @@
...
+      specify { assert_equal(File.join(directory, 'puppet.conf'), puppet_configurer.config_path) }
+      specify { assert_equal(File.join(directory, 'hiera.yaml'), puppet_configurer['hiera_config']) }
+      specify { assert File.file?(puppet_configurer['hiera_config']) }
+      specify { assert_equal(File.join(directory, 'environments'), puppet_configurer['environmentpath']) }
+      specify { assert_equal(File.join(directory, 'facts'), puppet_configurer['factpath']) }
+      specify { assert File.directory?(puppet_configurer['factpath']) }
+
+      it 'writes the correct facts' do
+        execution_environment.configure_puppet
+        facts = YAML.safe_load(File.read(File.join(directory, 'facts', 'kafo.yaml')))
```
mmoll: breaks on Ruby <2.1

ekohl:
Updated to use `YAML.load_file` where possible and otherwise use `YAML.load` instead of `YAML.safe_load`. This restores compatibility with Ruby 2.0 and removes a `File.read`.

Also introduced 2fb4afe103e738e534973df6e93191231b6e66a0 but this is only a partial deprecation fix. Also introduced 0e812d2f77c0a3916675d5ea1966c7559da78b51 for a more consistent scenario ID handling. 7cab6e9b78dadf398198d717fb308aa2b8faa80e now uses the `Configuration.scenario_id` property rather than copying the logic.

ekohl:
So one test was failed because it aborted after 60 minutes. I don't know why it was so slow.

lib/kafo/configuration.rb
```diff
@@ -47,6 +52,7 @@ def initialize(file, persist = true)
       end
 
       @config_dir = File.dirname(@config_file)
+      @scenario_id = Configuration::get_scenario_id(@config_file)
```
ares: I guess you meant `Configuration.get_scenario_id`, it's regular class method

test/kafo/configuration_test.rb
```diff
@@ -20,6 +20,12 @@ module Kafo
     specify { File.exist?(File.expand_path(basic_config.gem_root)) }
     specify { File.expand_path(basic_config.kafo_modules_dir).must_match %r|/modules$|}
 
+    describe '::get_scenario_id' do
```
ares: Replace `::` with `.` in here and next 4 lines

lib/kafo/scenario_manager.rb
```diff
@@ -19,7 +19,7 @@ def available_scenarios
           content = YAML.load_file(scn_file)
           if content.is_a?(Hash) && content.has_key?(:answer_file) && content.fetch(:enabled, true)
             # add scenario name for legacy configs
-            content[:name] = File.basename(scn_file, '.yaml') unless content.has_key?(:name)
+            content[:name] = Configuration::get_scenario_id(scn_file) unless content.has_key?(:name)
```
ares: Replace as elsewhere

lib/kafo/configuration.rb
ekohl: Isn't this the way to call it statically?

ekohl: Updated

ares:
Thanks @ekohl, test failure on one combination timedout, all the rest is green, merging!

PR #242, opened 2019-09-06
Allow creating the Hiera config in a subdirectory + typo fix

ekohl:
These are compatible changes from https://github.com/theforeman/kafo/pull/241 which I'm submitting in parts to see where the failure comes from.

### DIFF
--- a/lib/kafo/exit_handler.rb
+++ b/lib/kafo/exit_handler.rb
@@ -1,11 +1,10 @@
 module Kafo
   class ExitHandler
-    attr_accessor :cleanup_paths, :exit_code, :logger
+    attr_accessor :cleanup_paths, :exit_code
 
     def initialize
       @cleanup_paths = []
       @exit_code = 0
-      @logger = KafoConfigure.logger
     end
 
     def error_codes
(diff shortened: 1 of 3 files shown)


ekohl:
[test kafo]

lib/kafo/exit_handler.rb
```diff
@@ -1,11 +1,10 @@
 module Kafo
   class ExitHandler
-    attr_accessor :cleanup_paths, :exit_code, :logger
+    attr_accessor :cleanup_paths, :exit_code
```
ares: it may be better to keep logger as a writer and add customer reader

```
def logger
  @logger ||= KafoConfigure.logger
end
```
this way we also don't require KafoConfigure to be ready during initialization but with the first use of local logger

the benefit is, logger can still be changed to another instance if needed

ekohl: I also considered that, but was confused by https://github.com/theforeman/kafo/blob/afc25fe69570db9dd6ad8abd232e7bdfde359ebb/lib/kafo/exit_handler.rb#L30-L31

Should that also use the logger?

ares: probably yes

ekohl: My issue is that this is all highly inconsistent:
https://github.com/theforeman/kafo/blob/afc25fe69570db9dd6ad8abd232e7bdfde359ebb/lib/kafo/hooking.rb#L22-L24
https://github.com/theforeman/kafo/blob/afc25fe69570db9dd6ad8abd232e7bdfde359ebb/lib/kafo/migrations.rb#L23
https://github.com/theforeman/kafo/blob/afc25fe69570db9dd6ad8abd232e7bdfde359ebb/lib/kafo/puppet_module.rb#L16
https://github.com/theforeman/kafo/blob/afc25fe69570db9dd6ad8abd232e7bdfde359ebb/lib/kafo/puppet_module.rb#L45

For now I'll just fix this class.

ekohl:
Closing now that the minitest reporter gem is used which reports the actual class name.

PR #243, opened 2019-09-23
Fixes #27909 - Translate class names to files

ekohl:
Given the following mapping:

```yaml
:mapping:
 :foreman_proxy_certs:
   :manifest_name: foreman_proxy_content
   :dir_name: certs
   :params_name: foreman_proxy_content::params
```

This results in:

```
checking presence of /usr/share/foreman-installer/modules/certs/manifests/foreman_proxy_content::params.pp
```

With this patch it correctly translates the class name into a loadable path.

### DIFF
--- a/lib/kafo/puppet_module.rb
+++ b/lib/kafo/puppet_module.rb
@@ -160,7 +160,7 @@ module Kafo
     end
 
     def default_params_path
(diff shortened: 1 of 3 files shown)


ehelms:
[test kafo]

PR #244, opened 2019-10-11
Switch from ci_reporter_minitest to minitest-reporters

ekohl:
The last commit to ci_reporter_minitest was in 2014 and started to generate deprecation warnings. The minitest-reporters project is maintained.


ekohl:
The report does appear to be easier to read since it understands class names. One task aborted but I think that timeout is unrelated.

PR #248, opened 2020-01-21
Fixes #28827 - Use the correct progress_bar variable

ekohl:
29999419d8ef5288c3a87708e3cbabd896422576 introduced the execution environment but used the incorrect variable. Since Ruby doesn't fail on undefined variables, this silently failed.


sthirugn:
Tested patch in downstream 6.7 build and it works great. @ekohl thank you so much

ekohl:
Looks like the build failed on timeouts which happens sometimes due to some facts on some slaves. Haven't been able to pin it down, but since @sthirugn confirmed it works I'm merging this now.

PR #249, opened 2020-01-21
Use Puppet 6.6 evaltrace for progress

ekohl:
9fc5fd3c41acfdd01ffbc6c929d00987691da5d4 disabled the progress bar monkey patch on Puppet 6. Puppet 6.6.0 started to mention the progress when evaluating resources which can be used to update the progress bar.

This is currently a draft since the actual line message still needs to be extracted and causes the tests to fail.


ehelms:
Whats the check that ensures this still works with Puppet 5?

ekohl:
I should have provided https://tickets.puppetlabs.com/browse/PUP-6465 as context. That links to https://github.com/puppetlabs/puppet/commit/dd0157380dc45894532934147afd423601bc455d. As you can see, `Starting to evaluate the resource` changed to `Starting to evaluate the resource (%{progress} of %{total})`. I probably won't have time to finish this soon, but wanted a quick PoC to see the feasibility of it for a presentation.

PR #250, opened 2020-03-13
Refs #28799: Allow calling puppet from inside SCL context

ehelms:


ehelms:
This did require some updates to testing for this new method of shelling out but it is also updated to support newer Ruby and bundler where the creation of the new bundler environment inside a tmp directory can interfere with the calling environment.

lib/kafo/base_context.rb
```diff
@@ -24,12 +26,18 @@ def self.clear_caches
...
+        symbolize(JSON.load(result) || {})
       end
     end
 
     def self.facter_path
       @facter_path ||= PuppetCommand.search_puppet_path('facter')
     end
+
+    def self.run_command(command)
+      stdout, stderr, _status = Open3.capture3(command)
```
ekohl: Looking ahead, Facter 4 will be written in Ruby. Should we be future proof and clean the env already?

ehelms: I just forgot to put this back due to the testing I had to do to debug some issues.

tasks/jenkins.rake
```diff
@@ -1,5 +1,5 @@
 namespace :jenkins do
-  task :unit => ['jenkins:setup', 'rake:test']
+  task :unit => ['jenkins:setup', 'test:unit', 'test:puppet_modules', 'test:acceptance']
```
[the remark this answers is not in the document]
ehelms: Correct. Previously it would cause all of the prep work to run for all types of tests which can mask what stage actually fails inside Jenkins.

lib/kafo/puppet_command.rb
```diff
@@ -35,12 +35,28 @@ def append(suffix)
     end
 
     def self.search_puppet_path(bin_name)
-      bin_path = (::ENV['PATH'].split(File::PATH_SEPARATOR) + ['/opt/puppetlabs/bin']).find do |path|
+      #bin_path = (['/opt/puppetlabs/bin'] + ::ENV['PATH'].split(File::PATH_SEPARATOR)).find do |path|
```
ekohl: Pushed the debug version?

ehelms: Yep, testing your theory out.

lib/kafo/puppet_command.rb
```diff
@@ -35,12 +35,27 @@ def append(suffix)
     end
 
     def self.search_puppet_path(bin_name)
-      bin_path = (::ENV['PATH'].split(File::PATH_SEPARATOR) + ['/opt/puppetlabs/bin']).find do |path|
+      bin_path = (['/opt/puppetlabs/bin'] + ::ENV['PATH'].split(File::PATH_SEPARATOR)).find do |path|
```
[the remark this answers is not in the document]
ehelms: The tricky part is that in a runtime environment the opposite is true. This code will find and use the facter from the SCL and then break.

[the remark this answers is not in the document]
ehelms: I wonder if I am going to need a check if its in the test environment to decide which order?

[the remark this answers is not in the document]
ehelms: Yea, the tests seem to indicate this was as you predicted. But how to tell the difference btween the two environments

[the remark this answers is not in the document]
ekohl: I've been thinking about and having some issues with this. We may get away by checking if `puppet` is from the puppetlabs env and find facter with `/opt/puppetlabs/bin` as the preferred path. Not pretty, but it allows us to package Puppet if we want or use a system version (and use facter from the same packaging), but if AIO is used then facter should preferably also come from AIO.

lib/kafo/puppet_command.rb
```diff
@@ -36,11 +36,26 @@ def append(suffix)
 
     def self.search_puppet_path(bin_name)
       bin_path = (::ENV['PATH'].split(File::PATH_SEPARATOR) + ['/opt/puppetlabs/bin']).find do |path|
-        File.executable?(File.join(path, bin_name))
+        File.executable?(File.join(path, 'puppet'))
```
ekohl: That's actually not as ugly as I thought, but I think this deserves a comment :)

lib/kafo/puppet_command.rb
```diff
@@ -35,12 +35,32 @@ def append(suffix)
     end
 
     def self.search_puppet_path(bin_name)
+      # Find the location of the puppet executable and use that to
+      # determine the path of all executables
       bin_path = (::ENV['PATH'].split(File::PATH_SEPARATOR) + ['/opt/puppetlabs/bin']).find do |path|
-        File.executable?(File.join(path, bin_name))
+        File.executable?(File.join(path, 'puppet'))
```
ekohl: This should probably also check if facter is really there
```suggestion
        File.executable?(File.join(path, 'puppet')) && File.executable?(File.join(path, bin_name))
```

Then if bin_path is empty I think it should fall back to the regular search.

ehelms:
Updated based on comments, worked the kinks out with the CI testing.

test/acceptance/kafo_export_params_test.rb
```diff
@@ -4,7 +4,10 @@ module Kafo
   describe 'kafo-export-params' do
     describe '--help' do
       it 'outputs usage' do
-        code, out, _ = run_command('bin/kafo-export-params --help', :dir => Dir.pwd)
+        generate_installer
```
ekohl, line 7: I think this should be a `before` block, similar to how it's done in `describe '-f'`. The `md` format can also benefit from it so you might be able to move it all the way to the top so every test in this file first generates an installer and adds a manifest.

PR #255, opened 2020-06-16
Refs #29802 - replace --skip-checks-i-know-better with --disable-syst…

wbclark:
…em-checks

This change is the first gradual step in a larger series of changes aimed at consolidating checks into hooks.

### DIFF
--- a/lib/kafo/hook_context.rb
+++ b/lib/kafo/hook_context.rb
@@ -114,6 +114,11 @@ module Kafo
       !mod.nil?
     end
 
+    # Check if --disable-system-checks was passed by the user
+    def disable_system_checks?
+      self.kafo.disable_system_checks?
+    end
+
     # You can trigger installer exit by this method. You must specify exit code as a first
     # argument. You can also specify a symbol alias which is built-in (see exit_handler.rb
     # for more details).
--- a/lib/kafo/kafo_configure.rb
+++ b/lib/kafo/kafo_configure.rb
@@ -189,7 +189,12 @@ module Kafo
         @progress_bar = config.app[:colors] ? ProgressBars::Colored.new : ProgressBars::BlackWhite.new
       end
 
-      unless skip_checks_i_know_better?
+      if skip_checks_i_know_better?
+        STDERR.puts '--skip-checks-i-know-better is deprecated and will be removed. Please use --disable-system-checks'
+        logger.warn '--skip-checks-i-know-better is deprecated and will be removed. Please use --disable-system-checks'
+      end
+
+      unless disable_system_checks? || skip_checks_i_know_better?
         unless SystemChecker.check
           puts "Your system does not meet configuration criteria"
           self.class.exit(:invalid_system)
--- a/README.md
+++ b/README.md
@@ -975,7 +975,7 @@ exit $?
 ```
 
 If you want to ignore results of the check scripts, you can use the builtin
-parameter `--skip-checks-i-know-better` (or `-s`). This will completely
+parameter `--disable-system-checks` (or `-s`). This will completely
 disable running all system check scripts. Note that this option is
 not persisted between runs.
 


mmoll:
[test kafo]

ehelms:
Looks good. An additional idea could be to add a function to the hook context (https://github.com/theforeman/kafo/blob/master/lib/kafo/hook_context.rb) so that instead of calling `kafo.disable_system_check?` in a hook, the code can just call `disable_system_check?`

ehelms:
We could include a hidden option (https://github.com/mdub/clamp#hidden-options) that keeps around --skip-checks-i-know-better with a deprecation warning message when used that does what --disable-system-checks does today. This would allow us a transition period for users of this flag and less turbulence. Thoughts?

wbclark:
Keeping --skip-checks-i-know-better as a hidden option that prints and logs a deprecation message.

lib/kafo/kafo_configure.rb
```diff
@@ -129,7 +129,12 @@ def execute
         @progress_bar = self.class.config.app[:colors] ? ProgressBars::Colored.new : ProgressBars::BlackWhite.new
       end
 
-      unless skip_checks_i_know_better?
+      if skip_checks_i_know_better?
+        puts '--skip-checks-i-know-better is deprecated and will be removed. Please use --disable-system-checks'
```
ekohl, line 133: Can we write this to stderr?

lib/kafo/hook_context.rb
```diff
@@ -65,6 +65,11 @@ def module_enabled?(module_name)
       !mod.nil? && mod.enabled?
     end
 
+    # Check if --disable-system-checks was passed by the user
+    def disable_system_checks?
```
ekohl, line 69: Currently nothing reads `skip_checks_i_know_better?` so that ends up being ignored in most cases.

ekohl:
Thinking more about it: why would they be named system checks. There is also a hook that adds `--certs-skip-check`. Could that be unified in a check as well? I wouldn't consider that a system check.

So as an end user how would it feel if you could use:
```
foreman-installer --skip-checks
foreman-installer --skip-checks=all
foreman-installer --skip-checks=cpu,memory,certs
```

Not sure about the first and if that's possible, but the value `all` and `none` are possible.

For the boot context, you want something like the following hook:
```ruby
def register_check(name, description = nil)
  # ...
end
```

Then in `hooks/boot/01-cpu.rb` you'll end up with:
```ruby
register_check :cpu, 'Verify sufficient CPU cores are available'
```

And `hooks/boot/02-certs.rb` you'll have:
```ruby
register_check :certs, 'Verify certificate files to be consistent and valid for at least a month'
```

Ideally these individual checks would show up in `--help`.

Currently in CI we run with `--disable-system-checks` but our certificate checks should execute just fine. AFAIK it's only memory that's lower than recommended.

I've only considered this from the implementation perspective and my requirements so please also think about it from the user perspective.

lib/kafo/hook_context.rb
wbclark: `skip_checks_i_know_better?` is read still at https://github.com/theforeman/kafo/pull/255/files#diff-657b82e4fb9c80f11badef608b14de10R137 which is the original behavior

my intention is to use `disable_system_checks?` helper in foreman-installer in place of the old `--disable-system-checks` hook (i.e. to disable other hooks)

as I have this currently written, the user would notice the following similarities/differences:

1. `--skip-checks-i-know-better` should behave exactly the same as before, skipping only the checks but not any hooks, with the caveat that it prints a deprecation warning

2. `--disable-system-checks` would not only disable the same hooks it already does, but now will disable the checks as well. since we are planning to merge the checks into hooks, this feels like the correct behavior for the new combined option

I could have this helper return true for `--skip-checks-i-know-better` as well but i'd rather just have it continue the old behavior and print the deprecation warning which gives the suggestion to use the new combined option. is this agreeable to you @ekohl ?

ekohl: Kafo is a library and every now and then we do get an issue reported so it's used outside of foreman-installer as well. That means it needs to present a consistent UI/API.

I'd suggest there is just a single way to use checks as a consumer (mostly hooks, but also API). That's why I'm suggesting clear methods as hooks.

Next is the compatibility with older versions.

From an API perspective, right now there's `kafo.skip_checks_i_know_better?` that's callable in hooks (see `katello/hooks/pre_commit/05-puppet_certs_exist.rb` in current foreman-installer`). I think you can also check it via `app_value(:skip_checks_i_know_better)` but I don't see an example for that.

From the UI perspective there's `--skip-checks-i-know-better`.

You can choose the compatibility you want to keep. Since this likely ends up as a breaking change and a major version, IMHO it's fine to keep none. If needed, the consumers of kafo can implement compatibility via hooks. You can even provide an upgrade example in docs. The benefit is that you don't carry it in kafo.

If you want to keep UI compatibility, you can catch `--skip-checks-i-know-better` and trigger skipping all checks.

API compatibility is also possible, but I think it'd end up ugly, complex and of minor value.

ehelms: I originally suggested to @wbclark to keep `--skip-checks-i-know-better` compatibility for an amount of time in order to keep that for the community used to it. If we think that this is a minor thing and it's easier to drop it and simply communicate to users to change to the new syntax then I'd be happy to drop that and simplify things for developers.

ehelms:
Your idea @ekohl is a good one. There is a potentially large community of users used to `--disable-system-checks` and I'd ask that we keep that as an alias to `--skip-checks=all` to provide compatibility for some time.

ekohl:
I thought about that as well. I'm not sure how feasible this is, but perhaps we can implement a hook method that you can set to modify which checks are skipped. Then Kafo doesn't have to know about `--disable-system-checks` and kafo consumers can use it to implement `--skip-checks-i-know-better` if they desire it as well.

wbclark:
@ekohl @ehelms To be honest, from the perspective of an user of the foreman-installer, out of the hundreds of users I have worked with, I don't think I've ever seen anyone need to skip checks. Generally the checks are there for a good reason -- for example something like `checks/dns.rb` shouldn't ever be skipped because the application won't work without name resolution. If you can think of a non-devel counter-example I'd like to hear it.

So it seems to me that this feature is only really useful for development environments, like our CI environment where the memory check would fail.

From the standpoint of other Kafo consumers, I'm not sure if hooks or checks are used more broadly or if skipping them is something that an end user should ever need to do.

lib/kafo/hook_context.rb
ehelms: So the conclusion I gathered from this discussion is for @wbclark to drop `skip-checks-i-know-better` and go with `disable-system-checks` or even just `disable-checks` since we will major version bump Kafo with this change anyway.

ekohl: Yes, but if we're going to rename, we should really provide a proper implementation. Just renaming from `--skip-checks-i-know-better` to `--skip-checks` provides no value IMHO. If desired, foreman-installer can already drop `--disable-system-checks` and standardize on the `--skip-checks-i-know-better`.

I really don't want to rename things twice and there's no rush to IMHO.

ares:
While I appreciate all great suggestions in this PR, I'm sad that 7 months after this PR was opened, today I found myself struggling to find which option to use to skip the memory check. I think merging the first or the second version would have brought a value to users already, 7 months ago. Instead, we ended up describing the ultimately great functionality, but merging nothing.

I must say I agree with @wbclark that users generally don't and shouldn't skip checks. The reason for a bit unproffesional name "skip checks I know better" back then was to discourage users from using it. The reason it was added in the first place was for the case we mess up something in the first versions of the installer. We could tell users how to ignore and proceed. It saved us many times when we debugged fqdn check. But those times are luckilly gone.

Specifying what checks should be skipped is perhaps nice to have, but I haven't seen any user request for that. Also keeping backwards compatibility is nice, but after years I am not aware of any active user of it besides Foreman installers.

I think the pragmatic and reasonable approach here is to rename `skip checks I know better` to `disable system checks`. That's what it was meant for. If we ever clean up other checks, we can add `disable certs checks` and others. Before we get there, I'm sure we'd improve the user experience immediatelly and for a long time by merging the simple version.

ehelms:
A thought to move us forward and allow things to get completed in pieces. Taking @ekohl design idea, recapping it here:

```
foreman-installer --skip-checks
foreman-installer --skip-checks=all
foreman-installer --skip-checks=cpu,memory,certs
```

We could introduce `--skip-checks` where it *only* works initially as an *all* skip the way that both `--disable-system-checks` and `--skip-checks-i-know-better` work today. And deprecate both `--disable-system-checks` and `--skip-checks-i-know-better`. That begins the journey of users updating to `--skip-checks` and allows to introduce as an enhancement the ability to start passing specific checks to `--skip-checks` without having to commit to the entire re-design and re-factoring of the code up front.

ares:
Sounds reasonable to me if it moves us forward. Given there was no development on this for almost a year, I'd re-consider merging as is as it is already better than what we have today.

ehelms:
Replaced by https://github.com/theforeman/foreman-installer/pull/791

PR #256, opened 2020-06-25
Fixes #30218 - Handle arguments with -S in them

ekohl:
Previously the code to detect the -S parameter didn't check for the beginning. That meant that if --certificate A-ST.pem would be interpreted as --certificate A -ST.pem.

### DIFF
--- a/lib/kafo/scenario_manager.rb
+++ b/lib/kafo/scenario_manager.rb
@@ -78,12 +78,19 @@ module Kafo
 
     def scenario_from_args(arg_name='--scenario|-S')
       # try scenario provided in the args via -S or --scenario
-      parsed = ARGV.join(" ").match(/(#{arg_name})(\s+|[=]?)(\S+)/)
-      if parsed
-        scenario_file = File.join(config_dir, "#{parsed[3]}.yaml")
-        return scenario_file if File.exist?(scenario_file)
-        fail_now("Scenario (#{scenario_file}) was not found, can not continue", :unset_scenario)
+      ARGV.each_with_index do |arg, index|
+        parsed = arg.match(/^(#{arg_name})($|=(?<scenario>\S+))/)
+        if parsed
+          scenario = parsed[:scenario] || ARGV[index + 1]
+          next unless scenario
+
+          scenario_file = File.join(config_dir, "#{scenario}.yaml")
+          return scenario_file if File.exist?(scenario_file)
+          fail_now("Scenario (#{scenario_file}) was not found, can not continue", :unset_scenario)
+        end
       end
+
+      nil
     end
 
     def select_scenario
--- a/test/kafo/scenario_manager_test.rb
+++ b/test/kafo/scenario_manager_test.rb
@@ -182,6 +182,48 @@ module Kafo
       end
     end
 
+    describe '#scenario_from_args' do
+      before do
+        @old_argv = ::ARGV.dup
+        ::ARGV.clear
+        ::ARGV.concat(args)
+      end
+      after do
+        ::ARGV.clear
+        ::ARGV.concat(@old_argv)
+      end
+
+      describe 'with -S inside a word' do
+        let(:args) { ['--server-ca', 'A-ST.pem'] }
+
+        it 'ignores -S inside words' do
+          manager.stub(:fail_now, Proc.new { |message, _| raise message }) do
+            assert_nil manager.scenario_from_args
+          end
+        end
+      end
+
+      [
+        ['-S', 'myscenario'],
+        ['-S=myscenario'],
+        ['--scenario', 'myscenario'],
+        ['--scenario=myscenario'],
+      ].each do |argsx|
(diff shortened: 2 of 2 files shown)


wbclark:
This works well in my testing.

```
irb(main):001:0> arg_name='--scenario|-S'
=> "--scenario|-S"
irb(main):002:0> ARGV = ['--scenario', 'katello', '--cert', 'A-ST.pem']
=> ["--scenario", "katello", "--cert", "A-ST.pem"]
irb(main):003:0> ARGV.each_with_index do |arg, index|
irb(main):004:1*   parsed = arg.match(/^(#{arg_name})($|=(?<scenario>\S+))/)
irb(main):005:1>   if parsed
irb(main):006:2>     scenario = parsed[:scenario] || ARGV[index + 1]
irb(main):007:2>     next unless scenario
irb(main):008:2>     puts "#{scenario}"
irb(main):009:2>   end
irb(main):010:1> end
katello
=> ["--scenario", "katello", "--cert", "A-ST.pem"]
```

I tried changing the order of the arguments and found it worked in that case as well.

```
irb(main):011:0> ARGV = ['--cert', 'A-ST.pem', '--scenario', 'katello']
=> ["--cert", "A-ST.pem", "--scenario", "katello"]
irb(main):012:0> ARGV.each_with_index do |arg, index|
irb(main):013:1*   parsed = arg.match(/^(#{arg_name})($|=(?<scenario>\S+))/)
irb(main):014:1>   if parsed
irb(main):015:2>     scenario = parsed[:scenario] || ARGV[index + 1]
irb(main):016:2>     next unless scenario
irb(main):017:2>     puts "{#scenario}"
irb(main):018:2>   end
irb(main):019:1> end
katello
=> ["--cert", "A-ST.pem", "--scenario", "katello"]
```

There is one part I don't quite follow. I'll comment inline and would appreciate it if you would take a moment to explain.

lib/kafo/scenario_manager.rb
```diff
@@ -78,12 +78,19 @@ def configured?
...
       # try scenario provided in the args via -S or --scenario
-      parsed = ARGV.join(" ").match(/(#{arg_name})(\s+|[=]?)(\S+)/)
-      if parsed
-        scenario_file = File.join(config_dir, "#{parsed[3]}.yaml")
-        return scenario_file if File.exist?(scenario_file)
-        fail_now("Scenario (#{scenario_file}) was not found, can not continue", :unset_scenario)
+      ARGV.each_with_index do |arg, index|
+        parsed = arg.match(/^(#{arg_name})($|=(?<scenario>\S+))/)
+        if parsed
+          scenario = parsed[:scenario] || ARGV[index + 1]
```
wbclark, line 84: In the example I did with `ARGV = ["--scenario", "katello", "--cert", "A-ST.pem"]`, I found that the values of parsed were [nil, nil, #<MatchData "--scenario" scenario:nil>, nil] respectively.

Could you explain what `parsed[:scenario]` is doing here?

ekohl: The part `(?<scenario>\S+)` is a [named capture group](https://ruby-doc.org/core-2.7.1/Regexp.html#class-Regexp-label-Capturing). If you don't do that, you need to refer to it by number. Since a number can change is hard to follow, I chose a named group.

In the case of `['--scenario', 'katello']` there is no match so it's nil. That's why I do a look forward (`ARGV[index + 1]`) to get the next argument.

wbclark: Great, thanks for the reference

ehelms:
Is `-S` used anywhere? Could we drop support of the short option in favor of the explicit `--scenario` ? Given you only really need to pass `--scenario` once ever, having the short option does not save any users time and would prevent having to have a work around in our code.

wbclark:
> Is `-S` used anywhere? Could we drop support of the short option in favor of the explicit `--scenario` ? Given you only really need to pass `--scenario` once ever, having the short option does not save any users time and would prevent having to have a work around in our code.

I'd prefer not to remove the `-S` shorthand. It's fairly common for a lot of users, and I've seen some scripts that use it as well.

ekohl:
What value is there in removing it? The code isn't simplified by doing so and it creates churn. As @wbclark mentions, these end up in scripts or memory.

ehelms:
The value is no hack to deal with `-S` in our code or a special test to ensure we don't regress this odd situation. I don't see value in the short option for scenario. Users can easily adjust their scripts if they are specifying scenario every single time. 

Anyhoo, this was just a thought to avoid odd work arounds in our code.

ekohl:
> The value is no hack to deal with `-S` in our code or a special test to ensure we don't regress this odd situation. I don't see value in the short option for scenario. Users can easily adjust their scripts if they are specifying scenario every single time.

Looking at the code, it could have happened with `--server-cert my--scenario.pem` as well, it's just that `-S` happens to be the one we found. Given this has been there for years, it's clearly uncommon to have those in your arguments.

PR #257, opened 2020-07-02
Refs #29789 - add app_option? helper to hook extensions

wbclark:


lib/kafo/hook_context.rb
```diff
@@ -22,14 +22,20 @@ def logger
     end
 
     # if you want to add new app_option be sure to do as soon as possible (usually boot hook)
-    # otherwise it may be to late (e.g. when displaying help)
+    # otherwise it may be too late (e.g. when displaying help)
```
wbclark, line 25: It felt like overkill to create a 2nd PR or commit to fixup this comment. Since it's a related helper anyway, I included it here.

ekohl:
Could you take a stab at writing some tests for this?

ehelms:
I think a new Redmine issue under the Kafo project is needed for this: https://projects.theforeman.org/projects/kafo/issues

wbclark:
> Could you take a stab at writing some tests for this?

I think I have the write approach to test that `app_option?('unknown_option')` should return `false` when `unknown_option` doesn't exist.

I'm not sure how exactly to test for an option that does exist.

test/kafo/hook_context_test.rb
```diff
@@ -43,5 +44,17 @@ module Kafo
...
       end
     end
+
+    describe "#app_option?" do
+      specify do
+        app = {'known_option' => 'option_arg'}
+        config = Minitest::Mock.new
+        kafo.expect :config, config
+        config.expect :app, app
+        app.expect :key?['known_option'.to_sym]
```
ehelms, line 54: I think you might want `app.expect :key?, :known_option`

test/kafo/hook_context_test.rb
```diff
@@ -43,5 +44,17 @@ module Kafo
...
       end
     end
+
+    describe "#app_option?" do
+      specify do
+        app = {'known_option' => 'option_arg'}
+        config = Minitest::Mock.new
+        kafo.expect :config, config
+        config.expect :app, app
+        app.expect :key?, :known_option
```
ekohl, line 54: Why is this needed? You already have a hash with `known_option` so there's no need to mock anything

test/kafo/hook_context_test.rb
```diff
@@ -43,5 +44,16 @@ module Kafo
         assert_equal context.module_enabled?('known_module'), true
       end
     end
+
+    describe "#app_option?" do
+      specify do
+        app = {'known_option' => 'option_arg'}
```
ekohl, line 50: I suspect in reality the options will be symbols and that causes it to fail

test/kafo/hook_context_test.rb
```diff
@@ -43,5 +44,16 @@ module Kafo
...
+        config = Minitest::Mock.new
+        kafo.expect :config, config
+        config.expect :app, app
+        assert_equal context.app_option?('known_option'), true
```
ekohl, line 54: This is failing. It looks confusing because the [signature](https://docs.ruby-lang.org/en/2.1.0/MiniTest/Assertions.html#method-i-assert_equal) is `assert_equal(exp, act, msg = nil)`. That's why expected and actual are swapped if you look at https://ci.theforeman.org/job/kafo-pr-test/22/testReport/junit/(root)/Kafo__HookContext___app_option_/Test___Matrix___ruby____2_6___PUPPET_VERSION____5_0____test_0001_anonymous/

Looks like the  `#module_enabled?` test is doing the same thing.

test/kafo/hook_context_test.rb
```diff
@@ -43,5 +44,16 @@ module Kafo
...
+        assert_equal context.app_option?('unknown_option'), false
```
ekohl, line 55: Nit: In general it's considered a bad practice to combine multiple tests into a single `specify`. One is testing for the positive case, the other for the negative.

test/kafo/hook_context_test.rb
```diff
@@ -43,5 +44,20 @@ module Kafo
         assert_equal context.module_enabled?('known_module'), true
       end
     end
+
+    describe "#app_option?" do
+      app = {:known_option => 'option_arg'}
+      config = Minitest::Mock.new
+      kafo.expect :config, config
+      config.expect :app, app
```
ekohl, line 52: AFAIK this is generally a bad practice. The `specify` blocks scope variables to a test. The way you define it here might leak after the test. I'm not that familiar with minitest, but you may be able to use `before {}`.

ehelms: I will add, `expect` is a form of `assertion` and should be used within a `specify` block. You can use `let` to setup the mock itself in spec notation but don't be afraid to repeat yourself in tests -- they can be more verbose as tests should be easy to read and understand. Generally, you use `expect` if you care about the internals that are happening and want to assure they are called to some external API and stubs are used when you need a fake object but aren't trying to assert the internals. Since this test file uses Mocks instead of stubs extensively I think it's OK to continue that pattern.

```
let(:app) { {:known_option => 'option_arg'} }
let(:config) { Minitest::Mock.new }

specify do
  kafo.expect :config, config
  config.expect :app, app
  assert_equal true, context.app_option?('known_option')
end

**Comment by wbclark** [comment]
After running the tests locally,

```
Finished in 179.90667s
714 tests, 1093 assertions, 0 failures, 0 errors, 0 skips
```

I think this is good to go. Thanks @ehelms and @ekohl for the helpful reviews

ekohl:
[test kafo]

PR #259, opened 2020-07-10
Fixes #30365: Ignore symlinks for puppet executable

ehelms:


ekohl:
Looks like the tests fail on this

ehelms:
Yea, due to paths existing on local system I have to run these tests in the isolated Jenkins env >.<

ehelms:
Tests passing now

ehelms:
Related fix needed: https://github.com/theforeman/kafo_parsers/pull/40

ehelms:
@ekohl did you want to look at this for pulling into a release?

lib/kafo/puppet_command.rb
```diff
@@ -38,8 +38,11 @@ def self.search_puppet_path(bin_name)
       # Find the location of the puppet executable and use that to
       # determine the path of all executables
       bin_path = (::ENV['PATH'].split(File::PATH_SEPARATOR) + ['/opt/puppetlabs/bin']).find do |path|
```
ekohl, line 40: I still have doubts about this line. In production we actually want to prefer the last value over `$PATH` but in dev we don't. However, I think it's OK. The user with a non-trivial wrapper script could fix it and that's probably ok.

test/test_helper.rb
```diff
@@ -6,6 +6,7 @@
 require 'minitest/spec'
 require 'minitest/mock'
 require "minitest/reporters"
+require 'mocha/minitest'
```
ekohl, line 9: Why mocha? We now have 2 ways to stub/mock.

ehelms: Because minitest stubbing is insufficient to handle our use case.

ehelms:
@ekohl Anything else you need here to complete review?

test/test_helper.rb
ekohl: I'd prefer to get 4d26925692cda13fb52f0382686d0be72be6655c in to avoid the use of mocha.

ekohl:
Looks like CI is still running, but :+1: other than that.

PR #264, opened 2020-08-10
Revert "Fixes #30365: Ignore symlinks for puppet executable"

ehelms:
This reverts commit 695ad18f7f3dccb07d93604701f68345c665b181.


ehelms:
Turns out this breaks in production as `/opt/puppetlabs/bin` is nothing but symlinks. Somehow, I missed that.

PR #266, opened 2020-08-10
Fixes #30625 - add has_custom_fact? helper

wbclark:


lib/kafo/configuration.rb
```diff
@@ -104,6 +104,10 @@ def set_custom_fact(key, value)
       custom_fact_storage[key.to_s] = value
     end
 
+    def has_custom_fact?(key)
+      custom_fact_storage.key?(key.to_sym)
```
ehelms, line 108: This is turning `key` into a symbol, if I look at (https://github.com/theforeman/kafo/pull/266/files#diff-b3325340a3c63ce60780a66f5db027d5R100) that code is ensuring the key is a string using `to_s`.

wbclark: Thanks @ehelms . Updated and also added a first pass at a test case for the hook context.

test/kafo/hook_context_test.rb
```diff
@@ -85,5 +86,23 @@ module Kafo
...
       end
     end
+
+    describe "#has_custom_fact? with fact value present" do
+      let(:config) { Minitest::Mock.new }
+      let(:custom_fact_storage) { { 'my_custom_fact' => 'my_custom_fact_value' } }
+      kafo.expect :config, config
+      config.expect :custom_fact_storage, custom_fact_storage
+      assert_equal true, context.has_custom_fact?('my_custom_fact')
+      assert_equal false, context.has_custom_fact?('not_my_custom_fact')
```
ehelms, line 96: I think you want to wrap the 4 lines above this point in a `specify do` block. Same with the next test you added.

test/kafo/hook_context_test.rb
```diff
@@ -85,5 +86,29 @@ module Kafo
...
       end
     end
+
+    describe "#has_custom_fact? with fact value present" do
+      let(:config) { Minitest::Mock.new }
+      let(:custom_fact_storage) { { 'my_custom_fact' => 'my_custom_fact_value' } }
+
+      specify do
+        kafo.expect :config, config
+        config.expect :has_custom_fact?, custom_fact_storage
```
ehelms, line 96: I think based on the failure this needs to be:

```
config.expect :has_custom_fact?, custom_fact_storage, ['my_custom_fact']
```

test/kafo/hook_context_test.rb
```diff
@@ -85,5 +86,29 @@ module Kafo
...
+      end
+    end
+
+    describe "#has_custom_fact? with fact value nil" do
+      let(:config) { Minitest::Mock.new }
+      let(:custom_fact_storage) { { 'my_custom_fact' => nil } }
+
+      specify do
+        kafo.expect :config, config
+        config.expect :has_custom_fact?, custom_fact_storage
```
ehelms, line 108: See https://github.com/theforeman/kafo/pull/266/files#r471429365

test/kafo/hook_context_test.rb
wbclark: So if I understand this correctly, need to declare that the mocked function expects the fact name as an argument. I'll give this a try. Thanks @ehelms

wbclark:
Hmmm,

```
Failure:
test_0001_anonymous(Minitest::Result) [/home/jenkins/workspace/kafo-pr-test/test/kafo/hook_context_test.rb:109]:
Expected: true
  Actual: {"my_custom_fact"=>nil}
```

PR #272, opened 2020-08-27
Fixes #30723: Use Puppet 6.6 evaltrace for progress

ehelms:
9fc5fd3c41acfdd01ffbc6c929d00987691da5d4 disabled the progress bar
monkey patch on Puppet 6. Puppet 6.6.0 started to mention the progress
when evaluating resources which can be used to update the progress bar.

Replaces https://github.com/theforeman/kafo/pull/249

### DIFF
--- a/lib/kafo/progress_bar.rb
+++ b/lib/kafo/progress_bar.rb
@@ -10,9 +10,9 @@ module Kafo
 # #finite_template and #infinite_template methods. Also you may find useful to
 # change more methods like #done_message or #print_error
   class ProgressBar
-    MONITOR_RESOURCE = %r{\w*MONITOR_RESOURCE ([^\]]+\])}
-    EVALTRACE_START = %r{/(.+\]): Starting to evaluate the resource}
-    EVALTRACE_END = %r{/(.+\]): Evaluated in [\d\.]+ seconds}
+    MONITOR_RESOURCE = %r{\w*MONITOR_RESOURCE (?<resource>[^\]]+\])}
+    EVALTRACE_START = %r{/(?<resource>.+\]): Starting to evaluate the resource( \((?<count>\d+) of (?<total>\d+)\))?}
+    EVALTRACE_END = %r{/(?<resource>.+\]): Evaluated in [\d\.]+ seconds}
     PREFETCH = %r{Prefetching .* resources for}
 
     def initialize
@@ -37,12 +37,25 @@ module Kafo
       force_update = false
 
       if (line_monitor = MONITOR_RESOURCE.match(line))
-        @resources << line_monitor[1]
+        @resources << line_monitor[:resource]
         @total = (@total == :unknown ? 1 : @total + 1)
       end
 
       if (line_start = EVALTRACE_START.match(line))
-        if (known_resource = find_known_resource(line_start[1]))
+        if line_start[:total]
+          # Puppet 6.6 introduced progress in evaltrace
+          # Puppet counts 1-based where we count 0-based here
+          new_lines = line_start[:count].to_i - 1
+          new_total = line_start[:total].to_i
+          if new_lines != @lines || @total != new_total
+            @lines = new_lines
+            @total = new_total
+            update_bar = true
+            force_update = true
+          end
+        end
+
+        if (known_resource = find_resource(line_start[:resource]))
           line = known_resource
           update_bar = true
           force_update = true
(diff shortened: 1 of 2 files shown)


lib/kafo/progress_bar.rb
```diff
@@ -11,7 +11,7 @@ module Kafo
 # change more methods like #done_message or #print_error
   class ProgressBar
     MONITOR_RESOURCE = %r{\w*MONITOR_RESOURCE ([^\]]+\])}
-    EVALTRACE_START = %r{/(.+\]): Starting to evaluate the resource}
+    EVALTRACE_START = %r{/(.+\]): Starting to evaluate the resource( \((\d+) of (\d+)\))?}
```
ekohl, line 14: Back when I wrote the initial version we couldn't use capture groups in regexes because we needed to support Ruby 2.0. Now we can and that makes the code below a lot easier:
```ruby
regex = %r{/(?<resource>.+\]): Starting to evaluate the resource( \((?<number>\d+) of (?<total>\d+)\))?}
match = regex.match('/Stage[main]/Example/File[/foo/bar]: Starting to evaluate the resource (1 of 2)')
assert match[:resource] == 'Stage[main]/Example/File[/foo/bar]'
assert match[:number] == '1'
assert match[:total] == '2'

legacy_match = regex.match('/Stage[main]/Example/File[/foo/bar]: Starting to evaluate the resource')
assert legacy_match[:resource] == 'Stage[main]/Example/File[/foo/bar]'
assert legacy_match[:number].nil?
assert legacy_match[:total].nil?
```

lib/kafo/progress_bar.rb
```diff
@@ -42,15 +42,28 @@ def update(line)
...
+          new_total = line_start[4].to_i
+          if new_lines != @lines || @total != new_total
+            @lines = new_lines
+            @total = new_total
+            update_bar = true
+            force_update = true
+          end
+        end
+
+        if (known_resource = find_resource(line_start[1]))
```
ekohl, line 58: With Puppet 6, we no longer hook in internals. This part never runs:

https://github.com/theforeman/kafo/blob/6c7f49ff4e668853ca25ff07f1d4dd94ca7e9602/modules/kafo_configure/lib/puppet/parser/functions/add_progress.rb#L28

That means `MONITOR_RESOURCE` is a Puppet < 6 thing and `@resources` is never filled. So if you see `line_start[4]`, you know it's Puppet 6.6 and `find_known_resource` will never work.

This is the part I struggled with.

I think this should be an `elsif` and set `line` in the `if line_start[4]` part.

lib/kafo/progress_bar.rb
```diff
@@ -42,15 +42,28 @@ def update(line)
...
           force_update = true
         end
       end
 
       if (line_end = EVALTRACE_END.match(line)) && @total != :unknown && @lines < @total
-        if (known_resource = find_known_resource(line_end[1]))
+        if (known_resource = find_resource(line_end[1]))
           @resources.delete(known_resource)  # ensure it's only counted once
           @lines += 1
         end
```
ekohl, line 69: This feels wrong. It will always find something. I think that means it always increases `@lines` by 1, introducing an off-by-1 error.

lib/kafo/progress_bar.rb
ehelms: Switched for each resource to this method. This is much cleaner to read, thanks for pointing it out.

lib/kafo/progress_bar.rb
ehelms: Note how I changed this find_resource and re-factored the method to actually derive the resource. This does still feel nice so that you get the output as:

```
Notify[test]
Prefetching example resources for example_type 
File[/foo/bar]
```

Without deriving the resource as we had before I believe we end up with the reporting more like:

```
/Stage[main]/Example/Notify[test]
Prefetching example resources for example_type
/Stage[main]/Example/File[/foo/bar]
```

This is more verbose of output, and from a progress bar standpoint, I would think someone would only care to mentally parse, when reading, the way we do it today (e.g. File[/foo/bar]) rather than the full path. I think the other benefit is that it shows the most relevant info on the screen, for example:

```
Installing             Info: /Package[foreman-dynflow-sidekiq]: Starting  [6%] [.                       ]
```

Which, if it was the full path, I think would truncate the end off and user would not see the most important part. Granted, most of this information flies off the screen.

lib/kafo/progress_bar.rb
ehelms: Why will this always find something? If find_resource returns nil this won't trigger. I've not seen any issues testing this in practice.

PR #273, opened 2020-08-28
Split logging into a config class and instance class

ehelms:

PR #274, opened 2020-08-28
Use a logger to represent each stage 

ehelms:
Requires https://github.com/theforeman/kafo/pull/273

This includes a commit that moves to using a root logger with verbose mode as an appender. This simplifies the logic and structure a lot and is part of what enables having loggers per stage.

### DIFF
--- a/lib/kafo/hook_context.rb
+++ b/lib/kafo/hook_context.rb
@@ -3,14 +3,15 @@ require 'kafo/base_context'
 
 module Kafo
   class HookContext < BaseContext
-    attr_reader :kafo
+    attr_reader :kafo, :logger
 
-    def self.execute(kafo, &hook)
-      new(kafo).instance_eval(&hook)
+    def self.execute(kafo, logger, &hook)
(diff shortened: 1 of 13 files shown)


lib/kafo/hook_context.rb
```diff
@@ -3,22 +3,23 @@
...
+    attr_reader :kafo, :logger
 
-    def self.execute(kafo, &hook)
-      new(kafo).instance_eval(&hook)
+    def self.execute(kafo, logger, &hook)
+      new(kafo, logger).instance_eval(&hook)
     end
 
-    def initialize(kafo)
+    def initialize(kafo, logger)
```
wbclark, line 12: This appears to be breaking a lot of tests.

For example in `test/kafo/hook_context_test.rb:6`:

`   let(:context) { HookContext.new(kafo) }`

This now expects a 2nd argument

ehelms:
This is ready for a round of reviews

wbclark (approved):
Clear and functional. Tests look good. I like the root logger approach.

PR #275, opened 2020-09-04
Fixes #30781 - app_value() should expect Symbol

wbclark:


wbclark:
To be honest, this doesn't make a lot of sense to me.

My understanding is that Ruby best practices are not to enforce strict typing and instead rely on `respond_to?` rather than `is_a?` when necessary.

OK, best practices are just guidelines, but I don't see a strong reason to ignore them in this case. What is the real negative outcome if someone calls `app_value('my_option')` and `app_value` converts it to `:my_option` for us ?

ehelms:
I interpret it a bit more like a best practice enforcement, a linting type rule. If we *want* users to always pass symbols because they are more efficient and we want a consistent experience then we need to type enforce. This prevents having to remember they should be symbols when we review code as tests will simply fail. That is the value I saw.

ekohl:
One possible reason that it allowed strings was that Ruby < 2.2 had a memory leak if they were generated dynamically (according to https://stackoverflow.com/questions/16621073/when-to-use-symbols-instead-of-strings-in-ruby), though if `.to_sym` is called here I don't exactly know the benefit. Probably just adhering to the [Robustness Principle](https://en.wikipedia.org/wiki/Robustness_principle).

This is a breaking change and I don't like doing major version bumps all the time. At most this log a deprecation but even that I really doubt. We have seen that we have abysmal test coverage in our hooks so we're very likely to only hit this at runtime. That's why I think we should not do this. We should remember to use the best practices, but don't hard error on inefficiencies.

PR #276, opened 2020-09-10
Modify log format

ehelms:
Based on some of the feedback from [discourse discussion](https://community.theforeman.org/t/rfc-installer-output-and-logging-format/20053) I went with updating the log formatting to start with date to provide a consistent time series, followed by the log level and finally to encapsulate the stage in brackets to highlight it. This formatting is also consistent with other areas of the project such as Foreman's production log format.

Example output of new format:

```
2020-09-10 19:00:53 [ERROR] [configure] Could not find a suitable provider for augeas
2020-09-10 19:00:53 [ERROR] [configure] Could not find a suitable provider for foreman_config_entry
2020-09-10 19:00:53 [ERROR] [configure] Could not find a suitable provider for foreman_smartproxy
2020-09-10 19:00:53 [WARN]  [configure] Applied catalog in 10.28 seconds
2020-09-10 19:00:54 [INFO]  [configure] Puppet has finished, bye!
2020-09-10 19:00:54 [INFO]  [post] Executing hooks in group post
2020-09-10 19:00:54 [INFO]  [post] All hooks in group post finished
2020-09-10 19:00:54 [INFO]  [root] Installer finished in 29.181156873 seconds
```

This requires the work from: https://github.com/theforeman/kafo/pull/274

### DIFF
--- a/lib/kafo/logging.rb
+++ b/lib/kafo/logging.rb
@@ -6,30 +6,13 @@ require 'logging'
 module Kafo
   class Logging
 
-    PATTERN = "[%5l %d %c] %m\n"
-    ::Logging.color_scheme(
-      'bright',
-      :levels => {
-        :info  => :green,
-        :warn  => :yellow,
-        :error => :red,
-        :fatal => [:white, :on_red]
-      },
-      :date   => :blue,
-      :logger => :cyan,
-      :line   => :yellow,
-      :file   => :yellow,
-      :method => :yellow
-    )
-    COLOR_LAYOUT = ::Logging::Layouts::Pattern.new(:pattern => PATTERN, :color_scheme => 'bright')
-    NOCOLOR_LAYOUT = ::Logging::Layouts::Pattern.new(:pattern => PATTERN, :color_scheme => nil)
-
     class << self
       def root_logger
         @root_logger ||= ::Logging.logger.root
       end
 
       def setup(verbose: false)
+        set_color_scheme
         level = KafoConfigure.config.app[:log_level]
 
         setup_file_logging(
@@ -55,7 +38,7 @@ module Kafo
             'configure',
             :level => log_level,
             :filename => filename,
-            :layout => NOCOLOR_LAYOUT,
+            :layout => layout(false),
             :truncate => true
           )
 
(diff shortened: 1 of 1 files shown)


wbclark:
@ehelms since #274 is now merged, could you rebase this please?

ehelms:
Rebased

wbclark (commented):
Is there a particular reason to have brackets around the stage but not around the log level?

Something like this looks a little better to me visually:

```
2020-09-10 19:00:53 [ERROR] [configure] Could not find a suitable provider for augeas
2020-09-10 19:00:53 [ERROR] [configure] Could not find a suitable provider for foreman_config_entry
```
The added benefit is that as an user I can do something like:

`fgrep [ERROR] /var/log/foreman-installer/katello.log`

Without getting false matches like (excuse the old log format, I pulled this from a test system):

`[ WARN 2020-04-20T13:47:28 main]  # WARN, DEBUG, ERROR, FATAL, INFO, UNKNOWN`

ehelms:
I was mostly just trying to highlight the stage through it being in brackets. The arguments for being able to grep or parse the level of the message by wrapping it in brackets as well is a good one. I have updated this to reflect that and the example log output above, which would now look like:

```
2020-09-10 19:00:53 [ERROR] [configure] Could not find a suitable provider for augeas
2020-09-10 19:00:53 [ERROR] [configure] Could not find a suitable provider for foreman_config_entry
2020-09-10 19:00:53 [ERROR] [configure] Could not find a suitable provider for foreman_smartproxy
2020-09-10 19:00:53 [WARN]  [configure] Applied catalog in 10.28 seconds
2020-09-10 19:00:54 [INFO]  [configure] Puppet has finished, bye!
2020-09-10 19:00:54 [INFO]  [post] Executing hooks in group post
2020-09-10 19:00:54 [INFO]  [post] All hooks in group post finished
2020-09-10 19:00:54 [INFO]  [root] Installer finished in 29.181156873 seconds
```

lib/kafo/logging.rb
```diff
@@ -69,8 +52,33 @@ def setup_file_logging(log_level, log_dir, log_owner, log_group)
...
+          },
+          :date   => :blue,
+          :logger => :cyan,
+          :line   => :yellow,
+          :file   => :yellow,
+          :method => :yellow
+        )
+      end
+
+      def layout(color = false)
```
wbclark, line 72: Nit: I think that this is more generic than it needs to be, and the `color_layout` wrapper below is therefore not necessary. You could eliminate the wrapper by calling `KafoConfigure.use_colors?` directly inside `layout` instead of using the `colors` parameter, and then simply use `:layout => layout,` on L#89

wbclark (approved):
This looks good to me. I did leave one nit and I'll leave that up to you whether it's something you want to address before merging.

lib/kafo/logging.rb
ehelms: This is the spot that led to this design -- https://github.com/theforeman/kafo/pull/276/files#diff-d93616df475bf3004549b979a7abd583R41

I need to be able to explicitly get the layout without colors and be able to choose a color layout in the verbose case based on the configuration coming from Kafo.

ehelms: I will think on some further re-factoring to simplify this.

wbclark: OK, that makes sense. I took a stab at it with #279 to see what it might look like and you can let me know if you think it's worth doing.

PR #279, opened 2020-10-01
refactor logging color_layout

wbclark:


lib/kafo/logging.rb
```diff
@@ -36,10 +36,10 @@ def setup_file_logging(log_level, log_dir, log_owner, log_group)
         begin
           root_logger.appenders = ::Logging.appenders.rolling_file(
             'configure',
-            :level => log_level,
-            :filename => filename,
-            :layout => layout(false),
-            :truncate => true
+            level: log_level,
+            filename: filename,
+            layout: layout(color: false),
```
ehelms, line 41: If the default value is false we can skip passing this all together

wbclark: I would argue it's more readable this way, since it makes it clear we are using the nocolor layout without the reader having to find the method definition.

In that case we might as well drop the default value `false` instead, making `color:` a required keyword argument.

wbclark: Actually after looking at it, I do like keeping the default `color: false` which makes it clear we expect a boolean.

ehelms (approved):

PR #280, opened 2020-10-02
Label some options as advanced and hide by default

ehelms:
This introduces labeling some options as advanced and hiding them
under an Advanced section that is by default not displayed with
--help. These options are visible if a user supplies --full-help.
This aims to reduce the amount of options presented to a user when
using an installer based on Kafo at it's basic level.

This implementation also allows options created through hooks to label themselves as advanced

Example of change in the output after supplying `--full-help`:

```
= Generic:
    --reset-data                  This option will drop all databases for Foreman and subsequent backend systems. You will lose all data!
                                  Unfortunately, we can't detect a failure, so you should verify success manually.
                                  Dropping can fail when the DB is in use. (default: false)
    --detailed-exitcodes          Provide transaction information via exit codes, see puppet-agent(8)
                                  for full details. (default: false)
    -i, --interactive             Run in interactive mode
    --log-level LEVEL             Log level for log file output (default: :debug)
    -n, --noop                    Run puppet in noop mode? (default: false)
    -s, --skip-checks-i-know-better Skip all system checks (default: false)
    -v, --verbose                 Display log on STDOUT instead of progressbar
    -l, --verbose-log-level LEVEL Log level for verbose mode output (default: "info")
    -h, --help                    print help
    --full-help                   print complete help

= Advanced:
    --[no-]colors                 Use color output on STDOUT (default: true)
    --color-of-background COLOR   Your terminal background is :bright or :dark (default: :dark)
    --dont-save-answers           Skip saving answers to './config/foreman-answers.yaml'? (default: false)
    --ignore-undocumented         Ignore inconsistent parameter documentation (default: false)
    -p, --profile                 Run puppet in profile mode? (default: false)
    --skip-puppet-version-check   Skip check for compatible Puppet versions (default: false)
    --disable-scenario SCENARIO   Disable installation scenario
    --enable-scenario SCENARIO    Enable installation scenario
    --list-scenarios              List available installation scenarios
    --force                       Force change of installation scenario
    --compare-scenarios           Show changes between last used scenario and the scenario specified with -S or --scenario argument
    --migrations-only             Apply migrations to a selected scenario and exit
    --[no-]parser-cache           Force use or bypass of Puppet module parser cache
    --[no-]enable-foreman         Enable 'foreman' puppet module (default: true)
    --[no-]enable-foreman-cli     Enable 'foreman_cli' puppet module (default: true)
    --[no-]enable-foreman-cli-ansible Enable 'foreman_cli_ansible' puppet module (default: false)
    --[no-]enable-foreman-cli-azure Enable 'foreman_cli_azure' puppet module (default: false)
    --[no-]enable-foreman-cli-discovery Enable 'foreman_cli_discovery' puppet module (default: false)
    --[no-]enable-foreman-cli-kubevirt Enable 'foreman_cli_kubevirt' puppet module (default: false)
    --[no-]enable-foreman-cli-openscap Enable 'foreman_cli_openscap' puppet module (default: false)
    --[no-]enable-foreman-cli-remote-execution Enable 'foreman_cli_remote_execution' puppet module (default: false)
    --[no-]enable-foreman-cli-tasks Enable 'foreman_cli_tasks' puppet module (default: false)
    --[no-]enable-foreman-cli-templates Enable 'foreman_cli_templates' puppet module (default: false)
    --[no-]enable-foreman-proxy   Enable 'foreman_proxy' puppet module (default: true)
    --[no-]enable-puppet          Enable 'puppet' puppet module (default: true)
    --[no-]enable-foreman-plugin-ansible Enable 'foreman_plugin_ansible' puppet module (default: false)
    --[no-]enable-foreman-plugin-azure Enable 'foreman_plugin_azure' puppet module (default: false)
    --[no-]enable-foreman-plugin-bootdisk Enable 'foreman_plugin_bootdisk' puppet module (default: false)
    --[no-]enable-foreman-plugin-chef Enable 'foreman_plugin_chef' puppet module (default: false)
    --[no-]enable-foreman-plugin-default-hostgroup Enable 'foreman_plugin_default_hostgroup' puppet mod
```

### DIFF
--- a/lib/kafo/help_builders/base.rb
+++ b/lib/kafo/help_builders/base.rb
@@ -4,14 +4,17 @@ module Kafo
   module HelpBuilders
     DEFAULT_GROUP_NAME   = 'Basic'
     DEFAULT_MODULE_NAME  = 'Generic'
+    ADVANCED_MODULE_NAME = 'Advanced'
     IGNORE_IN_GROUP_NAME = /\s*parameters:?/
 
     class Base < Clamp::Help::Builder
       include StringHelper
 
-      def initialize(params)
+      def initialize(params, app_options: {}, advanced: false)
         super()
         @params = params
+        @app_options = app_options
+        @advanced = advanced
       end
 
       def add_list(heading, items)
@@ -23,6 +26,8 @@ module Kafo
           sorted_keys(data).each do |section|
             if section == 'Generic'
               add_list(header(1, section), data[section])
+            elsif section == 'Advanced'
+              add_list(header(1, section), data[section]) if @advanced
             else
               add_module(section, data[section])
             end
@@ -50,7 +55,7 @@ module Kafo
       end
 
       def module_header(name)
-        "\n" + header(1, 'Module ' + name)
+        header(1, 'Module ' + name)
       end
 
       def level(n)
@@ -73,7 +78,17 @@ module Kafo
       def by_module(help_items)
         data = Hash.new { |h, k| h[k] = [] }
         params_mapping(help_items).each do |item, param|
-          data[param.nil? ? DEFAULT_MODULE_NAME : param.module_name].push item
+          target = if param.nil?
+                     if @app_options.dig(item.attribute_name, :advanced)
+                       ADVANCED_MODULE_NAME
+                     else
+                       DEFAULT_MODULE_NAME
+                     end
+                   else
+                     param.module_name
+                   end
+
+          data[target].push(item)
         end
         data
       end
--- a/lib/kafo/kafo_configure.rb
+++ b/lib/kafo/kafo_configure.rb
@@ -187,14 +187,18 @@ module Kafo
     def self.help(*args)
       kafo          = args.pop
       builder_class = kafo.full_help? ? HelpBuilders::Advanced : HelpBuilders::Basic
-      args.push builder_class.new(kafo.params)
+      args.push builder_class.new(kafo.params, app_options: self.app_options, advanced: kafo.full_help?)
       super(*args)
     end
 
-    def self.app_option(*args, &block)
-      self.app_options ||= []
-      self.app_options.push self.option(*args, &block)
-      self.app_options.last
+    def self.app_option(switches, type, description, options ={}, &block)
+      self.app_options ||= {}
+
+      option = self.option(switches, type, description, options, &block)
+      advanced = options.dig(:advanced)
+
+      self.app_options[option.attribute_name] = {:option => option, :advanced => advanced}
+      self.app_options[option.attribute_name][:option]
     end
 
     def params
@@ -346,8 +350,8 @@ module Kafo
     # so we accept either allowed args or those that does not start with '-' and are right after
     # accepted argument
     def clamp_app_arguments
-      @allowed_clamp_app_arguments = self.class.app_options.map do |option|
-        option.switches.map { |s| is_yes_no_flag?(s) ? build_yes_no_variants(s) : s }
+      @allowed_clamp_app_arguments = self.class.app_options.map do |name, app_option|
+        app_option[:option].switches.map { |s| is_yes_no_flag?(s) ? build_yes_no_variants(s) : s }
       end
       @allowed_clamp_app_arguments.flatten!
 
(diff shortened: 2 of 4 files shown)


lib/kafo/kafo_configure.rb
```diff
@@ -293,33 +297,40 @@ def set_parameters
...
       self.class.app_option ['-s', '--skip-checks-i-know-better'], :flag, 'Skip all system checks', :default => false
-      self.class.app_option ['--skip-puppet-version-check'], :flag, 'Skip check for compatible Puppet versions', :default => false
+      self.class.app_option ['--skip-puppet-version-check'], :flag, 'Skip check for compatible Puppet versions', :default => false, :advanced => true
       self.class.app_option ['-v', '--verbose'], :flag, 'Display log on STDOUT instead of progressbar'
       self.class.app_option ['-l', '--verbose-log-level'], 'LEVEL', 'Log level for verbose mode output',
                             :default => 'info'
       self.class.app_option ['-S', '--scenario'], 'SCENARIO', 'Use installation scenario'
-      self.class.app_option ['--disable-scenario'], 'SCENARIO', 'Disable installation scenario'
-      self.class.app_option ['--enable-scenario'], 'SCENARIO', 'Enable installation scenario'
-      self.class.app_option ['--list-scenarios'], :flag, 'List available installation scenarios'
```
ekohl, line 318: I'm not sure this is advanced. If you consider `--scenario` as non-advanced, listing them doesn't feel advanced either.

lib/kafo/kafo_configure.rb
```diff
@@ -189,14 +189,18 @@ def help
...
     end
 
     def self.app_option(*args, &block)
-      self.app_options ||= []
-      self.app_options.push self.option(*args, &block)
-      self.app_options.last
+      self.app_options ||= {}
+      option = self.option(*args, &block)
+
+      advanced = args[3] ? args[3][:advanced] : false
```
ekohl, line 200: I never like these indexes. What is `3`? Isn't it cleaner to introduce `**kwargs` as a method parameter? Then advanced ends up in a that hash.

ekohl (commented):
Just some initial review. Overall I think the idea makes sense.

lib/kafo/kafo_configure.rb
ehelms: I struggled with these. I wanted to aim for a direction where `--help` showed the most useful, or common used set of parameters. And given we tend to pick a scenario and stick with it, I saw the scenario options as advanced. This may reduce visibility of scenarios, but for the predominant use cases this is something I'd rather revisit to focus the basic help.

ekohl: I agree that I debated them. However, I think that if you document `--scenario`, then `--list-scenarios` is also not advanced. enable/disable/force/compare are certainly advanced.

Btw, I don't know if we have the info already at this point, but perhaps we can be even smarter and see if there's more than one. If not, do not accept list options. You may want to always accept `--scenario` just so instructions like `foreman-installer --scenario foreman` does work, even if there is only one.

lib/kafo/help_builders/base.rb
```diff
@@ -4,14 +4,17 @@ module Kafo
...
     DEFAULT_GROUP_NAME   = 'Basic'
     DEFAULT_MODULE_NAME  = 'Generic'
+    ADVANCED_MODULE_NAME = 'Advanced'
     IGNORE_IN_GROUP_NAME = /\s*parameters:?/
 
     class Base < Clamp::Help::Builder
       include StringHelper
 
-      def initialize(params)
+      def initialize(params, app_options, advanced: false)
```
ekohl, line 13: Perhaps it should keep the API compatible?
```suggestion
      def initialize(params, app_options = {}, advanced: false)
```

lib/kafo/help_builders/basic.rb
```diff
@@ -15,7 +15,10 @@ def string
       private
 
       def except_resets(items)
-        items.select { |i| !i.help.first.strip.start_with?('--reset-') || !i.help.last.include?('to the default value (') }
+        items.select do |item|
+          !item.help.first.strip.start_with?('--reset-') ||
+            !item.help.last.include?('to the default value (')
+        end
```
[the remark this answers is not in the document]
ehelms: I'll just put it back.

lib/kafo/kafo_configure.rb
ehelms: I'm going to let #282 carry the conversation around the proposed change around hiding those options in some setups.

ehelms:
@ekohl I have split this into two commits, hoping that helps your ability to review the change

ekohl (commented):
I only looked at the first commit.

ehelms:
An alternative approach taking a more clamp native approach -- https://github.com/theforeman/kafo/pull/299

ekohl:
I like #299 a lot more - shall we close this?

PR #284, opened 2020-10-07
Add a puppet_execution_environment hook method

ekohl:
This makes it easy to run Puppet code in hooks. To achieve this, ExecutionEnvironment also gains `build_command`. Existing code is refactored to make use of this.

### DIFF
--- a/lib/kafo/execution_environment.rb
+++ b/lib/kafo/execution_environment.rb
@@ -42,6 +42,13 @@ module Kafo
       PuppetConfigurer.new(puppet_conf, settings)
     end
 
+    def build_command(code, options: [], settings: {}, use_answers: false)
+      store_answers if use_answers
+      puppetconf = configure_puppet(settings)
+      command = Kafo::PuppetCommand.new(code, options, puppetconf, @config).command
+      Kafo::PuppetCommand.format_command(command)
+    end
+
     private
 
     def environmentpath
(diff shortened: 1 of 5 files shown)


lib/kafo/execution_environment.rb
```diff
@@ -42,6 +42,13 @@ def configure_puppet(settings = {})
       PuppetConfigurer.new(puppet_conf, settings)
     end
 
+    def build_command(code, options: [], settings: {}, use_answers: false)
```
ehelms, line 45: I think I might like `build_puppet_command` to make it clearer to understand what sort of command is being built up. This method signature has the side affect of being able to be read as "build command" a noun rather than build as a verb.

ekohl: Alternative: would `command` make sense? You would get `execution_env.command(code)`.

ekohl:
Additional context: I had this in my mind when I wrote ExecutionEnvironment but couldn't find the right API so I left it out. I am thinking about dropping `PuppetCommand`.

For example, `append` on it is no longer used. If you drop the suffix, it essentially becomes a single method in the public interface.

lib/kafo/execution_environment.rb
ehelms: Can the command be anything I want? or is it specific to Puppet?

ehelms:
This would be another good one to get in.

PR #278, opened 2020-09-10
Add notice log level 

ehelms:
More often than not, the messages coming form puppet that has a level of warn is of the following:

```
[ WARN 2020-09-10T20:59:45 verbose] /Stage[main]/Foreman::Config/File[/etc/foreman/dynflow]: Skipping because of failed dependencies
[ WARN 2020-09-10T20:59:45 verbose] /Stage[main]/Foreman::Config/File[/etc/foreman/email.yaml]: Skipping because of failed dependencies
[ WARN 2020-09-10T20:59:45 verbose] /Stage[main]/Foreman::Config/File[/usr/share/foreman]: Skipping because of failed dependencies
[ WARN 2020-09-10T20:59:45 verbose] /Stage[main]/Foreman::Config/Group[foreman]: Skipping because of failed dependencies
```

In the context of the installer, this generally does not provide users with useful information as the cause of the skipped dependencies is due to an `ERROR` level message. This change moves the puppet warn / notice to a debug so that it is stilled captured in the log files as debug output. This aims to help `verbose` mode output the most pertinent information to a user when using `info` as the base level output.

PR #285, opened 2020-10-08
Switch to log style output as the default

ehelms:
Requires -- https://github.com/theforeman/kafo/pull/278

### DIFF
--- a/lib/kafo/kafo_configure.rb
+++ b/lib/kafo/kafo_configure.rb
@@ -68,7 +68,7 @@ module Kafo
       if ARGV.include?('--migrations-only')
         verbose = (ARGV.include?('--verbose') || ARGV.include?('-v'))
         Logging.setup(verbose: verbose)
-        self.class.logger.info('Log buffers flushed')
+        self.class.logger.notice('Log buffers flushed')
         self.class.exit(0)
       end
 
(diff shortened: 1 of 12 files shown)


lib/kafo/kafo_configure.rb
```diff
@@ -307,9 +311,9 @@ def set_app_options
                             :default => false
       self.class.app_option ['-s', '--skip-checks-i-know-better'], :flag, 'Skip all system checks', :default => false
       self.class.app_option ['--skip-puppet-version-check'], :flag, 'Skip check for compatible Puppet versions', :default => false
-      self.class.app_option ['-v', '--verbose'], :flag, 'Display log on STDOUT instead of progressbar'
```
wbclark, line 310: I think we could keep `--verbose`/`-v` as an option for a while, that simply prints a message that it is now the default behavior.

Otherwise it could be confusing to an user to try to use `--verbose` and receive only an error that the option doesn't exist.

lib/kafo/kafo_configure.rb
```diff
@@ -307,9 +311,9 @@ def set_app_options
...
-      self.class.app_option ['-l', '--verbose-log-level'], 'LEVEL', 'Log level for verbose mode output',
-                            :default => 'info'
+      self.class.app_option ['--progress-bar'], :flag, 'Display a progress bar for output', :default => false
+      self.class.app_option ['-l', '--verbose-log-level'], 'LEVEL', 'Log level for output',
```
wbclark, line 315:
```suggestion
      self.class.app_option ['-l', '--output-log-level'], 'LEVEL', 'Log level for output',
```
Or perhaps `--terminal-log-level`. The old naming no longer makes sense here.

PR #286, opened 2020-10-09
Add exit_code hook context method

ehelms:


lib/kafo/hook_context.rb
```diff
@@ -134,5 +134,10 @@ def scenario_path
     def scenario_data
       self.kafo.config.app
     end
+
+    # Return the current exit code
+    def exit_code
+      self.kafo.exit_code
+    end
```
wbclark, line 141: For my benefit, could you please confirm my understand of what this does?

1. Kafo initializes with an exit code of 0. At any point if an issue is encountered, that value can be changed

2. This method allows us to check the current value of the exit code within a hook

ehelms: That is correct. Today this is used by a post hook and checks the value directly e.g. `@kafo.exit_code` rather than using a method that is within the context itself, i.e. `exit_code`

ekohl (commented):
Is this really relevant? Currently it's only used in post hooks and exit_code is not even set before that so it'll not contain any relevant value.

Perhaps we should have a `PostHookContext` where it is available. Roughly

```ruby
module Kafo
  class PostHookContext < HookContext
    # Return the current exit code
    def exit_code
      self.kafo.exit_code
    end
  end
end
```

Similar, perhaps we need a `BootHookContext` since some methods may not be available at that stage.

An alternative to exposing the exit code is `apply_succeeded?` or `puppet_succeeded?` that has the logic as well to treat code 2 as a success.

ehelms:
You *could* set the exit code and continue on for hook code to pick it up later. The `foreman-installer` does not do this today, and in most cases this value will be the default 0.

This change is designed to go more cleanly with https://github.com/theforeman/kafo/pull/287 in which case the exit code could be relevant to both it and the post hook. I don't see that this hurts anything, and it provides a defined hook method rather than the way the exit code is retried today directly through the Kafo object.

`apply_succeeded?` works if we want a specific solution rather than a more generic solution that hooks can use.

wbclark (approved):
With #287 now merged and branching upon us, I'll merge this one as well to support a better implementation of https://github.com/theforeman/foreman-installer/pull/587

PR #287, opened 2020-10-10
Add exit hook

ehelms:
An additional hook is added that is executed during the exit
handler. This provides an opportunity to perform cleanup actions
or error messages that a user should see anytime the installer
exits for example.


ehelms:
Before:

```
2020-10-09 12:27:17 [NOTICE] [post] Executing hooks in group post
2020-10-09 12:27:17 [NOTICE] [post] All hooks in group post finished
```

After:

```
2020-10-10 19:15:47 [INFO ] [configure] Puppet has finished, bye!
2020-10-10 19:15:47 [INFO ] [post] Executing hooks in group post
2020-10-10 19:15:54 [INFO ] [post] All hooks in group post finished
```

lib/kafo/hooking.rb
```diff
@@ -44,15 +45,15 @@ def loaded?
       @loaded
     end
 
-    def execute(group)
+    def execute(group, skip_stage_log: false)
```
ekohl, line 48: I would use the positive to avoid double negatives.
```suggestion
    def execute(group, log_stage: true)
```

README.md
```diff
@@ -736,6 +736,7 @@ We currently support the following hooks.
 * pre_commit - after validations or interactive wizard have completed, all parameter values are set but not yet stored in the answer file
 * pre  - just before puppet is executed to converge system, after parameter values are stored in the answer file
 * post  - just after puppet is executed to converge system
+* exit  - happens as part of exit handling
```
ekohl, line 739: I'd consider `pre_exit` to be consistent with `pre_migrations`, `pre_values` and `pre_validations`.

ekohl (commented):
https://github.com/theforeman/kafo/blob/master/doc/kafo_run.uml should also be updated (and the PNG regenerated).

doc/kafo_run.uml
```diff
@@ -105,5 +105,7 @@
 
   #Greenyellow:**post**|
 
+  #Greenyellow:**pre_exit**|
```
ekohl, line 108: Should every hook have a line with `exit()` to `pre_exit`?

An example: let's say a `pre_commit` hook calls `exit()`, is `pre_exit` still called? (the code suggests it does, but the chart should document this I think).

ehelms: Anything that exits, or errors can lead to this occurring and calling the exit handler hook which would make for quite a messy diagram.

I can not seem to find what syntax this is using as it's quite old and plantuml has gone through changes. I also experimented with various attempts at this and none of them made it clearer.

lib/kafo/exit_handler.rb
```diff
@@ -26,6 +26,7 @@ def error_codes
     def exit(code, &block)
       @exit_code = translate_exit_code(code)
       block.call if block
+      KafoConfigure.hooking.execute(:pre_exit, log_stage: false)
```
ekohl, line 29: I must say that I dislike the hardcoded dependency on hooking here (just as I don't like the use of `KafoConfigure.logger`). IMHO it would be cleaner if `hooking` was passed in via the constructor as an optional parameter.

ekohl:
> `2020-10-09 12:27:17 [NOTICE] [post] Executing hooks in group post`

I think this part is an internal detail we shouldn't print at notice level.

ehelms:
> I think this part is an internal detail we shouldn't print at notice level.

The way I look at it, right now we show the following for a hook stage:

```
2020-10-08 00:02:31 [NOTICE] [init] Executing hooks in group init
2020-10-08 00:02:31 [NOTICE] [init] All hooks in group init finished
```

This shows the user that we entered a new stage, and then we finished the stage. I think those two elements of it are important at a notice level as we are wanting to show progression and stage.

lib/kafo/hooking.rb
```diff
@@ -10,7 +10,8 @@ class Hooking
     # pre_commit - after validations or interactive wizard have completed, all parameter values are set but not yet stored in the answer file
     # pre - just before puppet is executed to converge system
     # post - just after puppet is executed to converge system
```
wbclark, line 12: For the sake of consistency the new type of hook should have a comment here explaining how it is used, similar to the others.

lib/kafo/hooking.rb
```diff
@@ -44,15 +45,15 @@ def loaded?
...
       logger = Logger.new(group)
-      logger.info "Executing hooks in group #{group}"
+      logger.info "Executing hooks in group #{group}" if log_stage
       self.hooks[group].keys.sort_by(&:to_s).each do |name|
         hook = self.hooks[group][name]
         result = HookContext.execute(self.kafo, logger, &hook)
         logger.debug "Hook #{name} returned #{result.inspect}"
       end
-      logger.info "All hooks in group #{group} finished"
+      logger.info "All hooks in group #{group} finished" if log_stage
```
wbclark, line 56: With #278 now merged, this can now be updated to use the level `notice`. The same applies a few lines above.

wbclark (approved):
Perhaps it's not perfect but I think it's an improvement over where we are currently, and enables us to give clear and actionable feedback to user in case of installer failures.

PR #288, opened 2020-10-11
Implement multi hooks

ekohl:
Multi hooks are a special hook that allow you to define multiple hook stages within a single file. This can to avoid duplication of things like constants.

Currently still a draft since I haven't written documentation nor tests.

### DIFF
--- a/lib/kafo/hooking.rb
+++ b/lib/kafo/hooking.rb
@@ -1,4 +1,5 @@
 require 'kafo/hook_context'
+require 'kafo/multi_stage_hook'
 
 module Kafo
   class Hooking
@@ -35,6 +36,14 @@ module Kafo
             register(hook_type, file, &hook_block)
           end
         end
+
+        # Multi stage hooks are special
+        Dir.glob(File.join(base_dir, 'multi', '*.rb')).sort.each do |file|
+          logger.debug "Loading multi stage hook #{file}"
+          hook = File.read(file)
+          MultiStageHook.new(file, self, TYPES).instance_eval(hook, file, 1)
+        end
+
         @loaded = true
       end
       self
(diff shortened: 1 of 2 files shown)


ehelms:
I think an example would help understand the design and intent.

ekohl:
See https://github.com/theforeman/foreman-installer/pull/589 for a (working) example.

ehelms:
I like the grouping concept as it provides some end-to-end view of hooks with multiple parts that I have always wanted. This almost more so introduces a way to create something akin to a feature or parameter that uses the hook system to implement it. I am thinking on a term that spells that out more than multi and would strongly have us consider putting these in their own top-level directory to reduce confusion.

I think all told, I'd almost *rather* this method on the whole. You could define priorities and ordering directly which this current implementation lacks:

```
pre(10) do

end

post(15) do

end
```

ekohl:
> You could define priorities and ordering directly which this current implementation lacks:

It doesn't lack it - it uses the file name by default, but you can use a name (no example, not tested):
```ruby
pre(:15-my_hook) do
end
```

This can also be used to implement multiple hooks in the same stage (if needed).

ehelms:
Thinking more, a quantifier would help me:

 * multi stage hook
 * multi group hook

Also, adding this to the README :)

ekohl:
Exactly why it's a draft (due to the lack of a Proof Of Concept PR). I think multi stage hook feels better than group. Any thoughts about phase? Still leaning to stage.

ehelms:
Stage is always the word that pops into my head when I think about the installer and hooks. Stage has a flow feel to it, which hooks are part of, whereas group I agree does not make you think flow. Phase is synonymous but I think stage just is what my mind comes back with with installer.

ekohl:
Exactly my feelings about it.

ehelms:
I'd be happy to take this with the change to `multi_stage` to enable starting to consider building hooks this way.

ekohl:
For now I did rename it to MultiStageHook but I don't want to merge this without documentation and tests. Given the current time constraints I'll get back to it after branching.

ekohl:
https://github.com/theforeman/kafo/pull/333 was merged & released so this can be closed.

PR #289, opened 2020-10-13
Implement autoloading

ekohl:
Autoloading prevents having to think about what to exactly require. It can also postpone loading of classes until they're actually used. This can help with with application startup speed.

I only did minimal testing on this, but I think it makes the code cleaner. I considered [autoloaded](https://github.com/njonsson/autoloaded) but wanted to avoid an additional dependency.

### DIFF
--- a/lib/kafo.rb
+++ b/lib/kafo.rb
@@ -1,3 +1,50 @@
 # encoding: UTF-8
+
 require 'kafo/version'
-require 'kafo/kafo_configure'
+
+# First of all we have to store ENV variable, requiring facter can override them
+module Kafo
+  module ENV
+    LANG = ::ENV['LANG']
+  end
+
+  autoload :BaseContext, 'kafo/base_context'
+  autoload :ColorScheme, 'kafo/color_scheme'
+  autoload :Condition, 'kafo/condition'
+  autoload :ConditionError, 'kafo/exceptions'
+  autoload :Configuration, 'kafo/configuration'
+  autoload :ConfigurationException, 'kafo/exceptions'
+  autoload :DataType, 'kafo/data_type'
+  autoload :DataTypeParser, 'kafo/data_type_parser'
+  autoload :DataTypes, 'kafo/data_type' # This is inconsistent
+  autoload :ExecutionEnvironment, 'kafo/execution_environment'
+  autoload :ExitHandler, 'kafo/exit_handler'
+  autoload :FactWriter, 'kafo/fact_writer'
+  autoload :HelpBuilders, 'kafo/help_builders'
+  autoload :HieraConfigurer, 'kafo/hiera_configurer'
+  autoload :HookContext, 'kafo/hook_context'
+  autoload :Hooking, 'kafo/hooking'
+  autoload :KafoConfigure, 'kafo/kafo_configure'
+  autoload :Logger, 'kafo/logger'
+  autoload :Logging, 'kafo/logging'
+  autoload :MigrationContext, 'kafo/migration_context'
+  autoload :Migrations, 'kafo/migrations'
+  autoload :Param, 'kafo/param'
+  autoload :ParamBuilder, 'kafo/param_builder'
+  autoload :ParamGroup, 'kafo/param_group'
+  autoload :ParserCacheReader, 'kafo/parser_cache_reader'
+  autoload :ParserCacheWriter, 'kafo/parser_cache_writer'
+  autoload :ParserError, 'kafo/exceptions'
+  autoload :ProgressBar, 'kafo/progress_bar'
+  autoload :ProgressBars, 'kafo/progress_bars'
+  autoload :PuppetCommand, 'kafo/puppet_command'
+  autoload :PuppetConfigurer, 'kafo/puppet_configurer'
+  autoload :PuppetLogParser, 'kafo/puppet_log_parser'
+  autoload :PuppetModule, 'kafo/puppet_module'
+  autoload :ScenarioManager, 'kafo/scenario_manager'
+  autoload :Store, 'kafo/store'
+  autoload :StringHelper, 'kafo/string_helper'
+  autoload :SystemChecker, 'kafo/system_checker'
+  autoload :TypeError, 'kafo/exceptions'
+  autoload :Wizard, 'kafo/wizard'
+end
--- a/lib/kafo/kafo_configure.rb
+++ b/lib/kafo/kafo_configure.rb
@@ -1,6 +1,5 @@
 # encoding: UTF-8
 
-require 'pty'
 require 'clamp'
 
 module Kafo
@@ -420,6 +419,7 @@ module Kafo
         log_parser = PuppetLogParser.new
         logger = Logger.new('configure')
 
+        require 'pty'
         PTY.spawn(*PuppetCommand.format_command(command)) do |stdin, stdout, pid|
           begin
             stdin.each do |line|
--- a/lib/kafo/kafo_configure.rb
+++ b/lib/kafo/kafo_configure.rb
@@ -1,33 +1,7 @@
 # encoding: UTF-8
 
-# First of all we have to store ENV variable, requiring facter can override them
-module Kafo
-  module ENV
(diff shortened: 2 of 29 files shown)


ehelms:
> Autoloading prevents having to think about what to exactly require.

Why is this a good thing? Like Python, I think it's beneficial for developers to be able to see what is required within a file to be able to trace where functionality is coming from.

ekohl:
> Why is this a good thing? Like Python, I think it's beneficial for developers to be able to see what is required within a file to be able to trace where functionality is coming from.

With Ruby it doesn't really work that way. For example, Python has a fixed way how the import path works. With Ruby you can require any path and it can have a completely different content.

This was a bit inspired by Rails which uses autoloading (even though it's really a core Ruby feature). See https://edgeguides.rubyonrails.org/autoloading_and_reloading_constants.html as well. Note Foreman is not yet using Zeitwerk and is still on the classic loader.

My major annoyance is that you constantly need to be aware of what's loaded. This means a ton of duplication in tests. If you're loading the entire application, it can also slow things down. That can be seen in Rake tasks which load the world and then some.

Some unrepresentative testing. Without autoloading:

```console
$ time be ./bin/foreman-installer --list-scenarios
/home/ekohl/.gem/ruby/gems/highline-1.7.10/lib/highline.rb:624: warning: Using the last argument as keyword parameters is deprecated
Available scenarios
/home/ekohl/.gem/ruby/gems/highline-1.7.10/lib/highline.rb:624: warning: Using the last argument as keyword parameters is deprecated
  Foreman Proxy Content /home/ekohl/.gem/ruby/gems/highline-1.7.10/lib/highline.rb:624: warning: Using the last argument as keyword parameters is deprecated
(use: --scenario foreman-proxy-content)
/home/ekohl/.gem/ruby/gems/highline-1.7.10/lib/highline.rb:624: warning: Using the last argument as keyword parameters is deprecated
        Install a stand-alone Foreman Content Proxy.
/home/ekohl/.gem/ruby/gems/highline-1.7.10/lib/highline.rb:624: warning: Using the last argument as keyword parameters is deprecated
  Katello /home/ekohl/.gem/ruby/gems/highline-1.7.10/lib/highline.rb:624: warning: Using the last argument as keyword parameters is deprecated
(use: --scenario katello)
/home/ekohl/.gem/ruby/gems/highline-1.7.10/lib/highline.rb:624: warning: Using the last argument as keyword parameters is deprecated
        Install Foreman with Katello
/home/ekohl/.gem/ruby/gems/highline-1.7.10/lib/highline.rb:624: warning: Using the last argument as keyword parameters is deprecated
  Foreman /home/ekohl/.gem/ruby/gems/highline-1.7.10/lib/highline.rb:624: warning: Using the last argument as keyword parameters is deprecated
(use: --scenario foreman)
/home/ekohl/.gem/ruby/gems/highline-1.7.10/lib/highline.rb:624: warning: Using the last argument as keyword parameters is deprecated
        Default installation of Foreman

real	0m0.420s
user	0m0.361s
sys	0m0.056s
```

With autoloading:
```console
$ time be ./bin/foreman-installer --list-scenarios
/home/ekohl/.gem/ruby/gems/highline-1.7.10/lib/highline.rb:624: warning: Using the last argument as keyword parameters is deprecated
Available scenarios
/home/ekohl/.gem/ruby/gems/highline-1.7.10/lib/highline.rb:624: warning: Using the last argument as keyword parameters is deprecated
  Foreman Proxy Content /home/ekohl/.gem/ruby/gems/highline-1.7.10/lib/highline.rb:624: warning: Using the last argument as keyword parameters is deprecated
(use: --scenario foreman-proxy-content)
/home/ekohl/.gem/ruby/gems/highline-1.7.10/lib/highline.rb:624: warning: Using the last argument as keyword parameters is deprecated
        Install a stand-alone Foreman Content Proxy.
/home/ekohl/.gem/ruby/gems/highline-1.7.10/lib/highline.rb:624: warning: Using the last argument as keyword parameters is deprecated
  Katello /home/ekohl/.gem/ruby/gems/highline-1.7.10/lib/highline.rb:624: warning: Using the last argument as keyword parameters is deprecated
(use: --scenario katello)
/home/ekohl/.gem/ruby/gems/highline-1.7.10/lib/highline.rb:624: warning: Using the last argument as keyword parameters is deprecated
        Install Foreman with Katello
/home/ekohl/.gem/ruby/gems/highline-1.7.10/lib/highline.rb:624: warning: Using the last argument as keyword parameters is deprecated
  Foreman /home/ekohl/.gem/ruby/gems/highline-1.7.10/lib/highline.rb:624: warning: Using the last argument as keyword parameters is deprecated
(use: --scenario foreman)
/home/ekohl/.gem/ruby/gems/highline-1.7.10/lib/highline.rb:624: warning: Using the last argument as keyword parameters is deprecated
        Default installation of Foreman

real	0m0.381s
user	0m0.320s
sys	0m0.058s
```

So roughly a 10% improvement in start up time when you only want to list scenarios. Granted, it's all already sub-second, but the bigger the program becomes, the more you'll see it helps. For example, https://github.com/theforeman/kafo/pull/290 was one where I think that adding a class file and the autoloader definition is easier than thinking about where to add it in the loading. I also ran into similar problems with the Smart Proxy.

lib/kafo/kafo_configure.rb
```diff
@@ -446,6 +419,7 @@ def run_installation
         log_parser = PuppetLogParser.new
         logger = Logger.new('configure')
 
+        require 'pty'
```
ehelms, line 422: I dislike these mid-code requires. They can hide errors, and lead to missing libraries at runtime.

ehelms:
If I take #290  as the example, I can see that you introduced PuppetReport here -- https://github.com/theforeman/kafo/pull/290/files#diff-ef4f6d9e6e9975162ff8026e770e8787730a3547080b75534b8bdb411ff6efc2R478

And that by requiring that class here -- https://github.com/theforeman/kafo/pull/290/files#diff-ef4f6d9e6e9975162ff8026e770e8787730a3547080b75534b8bdb411ff6efc2R25

I now am able to trace where PuppetReport is coming from rather than having to guess if its from the code base itself, or some third party gem. I have been frustrated by this in the Ruby world many times. And Rails is the worst offender in this case.

ekohl:
This gem follows the convention that everything is in lib and then convert CamelCase into snake_case. So `Kafo::PuppetReport` can be found in `lib/kafo/puppet_report.rb`. So there is no need to look it up, the class name predicts it. https://github.com/njonsson/autoloaded automates this and has more about it but I didn't want to add a gem for it.

lib/kafo.rb
```diff
@@ -1,3 +1,50 @@
...
+  autoload :PuppetConfigurer, 'kafo/puppet_configurer'
+  autoload :PuppetLogParser, 'kafo/puppet_log_parser'
+  autoload :PuppetModule, 'kafo/puppet_module'
+  autoload :ScenarioManager, 'kafo/scenario_manager'
+  autoload :Store, 'kafo/store'
+  autoload :StringHelper, 'kafo/string_helper'
+  autoload :SystemChecker, 'kafo/system_checker'
+  autoload :TypeError, 'kafo/exceptions'
+  autoload :Wizard, 'kafo/wizard'
+end
```
ehelms, line 50: This seems overly verbose compared to something more automatic -- https://github.com/theforeman/tool_belt/blob/master/lib/tool_belt.rb#L2

ehelms:
> This gem follows the convention that everything is in lib and then convert CamelCase into snake_case. So `Kafo::PuppetReport` can be found in `lib/kafo/puppet_report.rb`. So there is no need to look it up, the class name predicts it. https://github.com/njonsson/autoloaded automates this and has more about it but I didn't want to add a gem for it.

Can this convention be enforced?

lib/kafo.rb
ekohl: The difference is that tool_belt is loading it all up front where `autoload` is lazy loading. You can also copy the pattern used by autoloaded to automatically convert file names to class/module names.

PR #290, opened 2020-10-13
Make Puppet reports available from the ExecEnv

ekohl:
Puppet produces reports. These can be used to produce good instructions to users about what exactly failed.

This is just the initial structure - it doesn't actually work yet and no useful information is retrieved from the actual report.

### DIFF
--- a/lib/kafo/execution_environment.rb
+++ b/lib/kafo/execution_environment.rb
@@ -19,6 +19,18 @@ module Kafo
       end
     end
 
+    def reportdir
+      @reportdir ||= begin
+        path = File.join(directory, 'reports')
+        File.mkdir(path)
+        path
+      end
+    end
+
+    def reports
+      Dir.entries(reportdir).sort_by { |path| File.mtime(path) }
+    end
+
     def store_answers
       answer_data = HieraConfigurer.generate_data(@config.modules, @config.app[:order])
       @logger.debug("Writing temporary answers to #{answer_file}")
@@ -37,6 +49,8 @@ module Kafo
         'environmentpath' => environmentpath,
         'factpath'        => factpath,
         'hiera_config'    => hiera_config,
+        'reports'         => 'store',
+        'reportdir'       => reportdir,
       }.merge(settings)
 
       PuppetConfigurer.new(puppet_conf, settings)
--- a/lib/kafo/kafo_configure.rb
+++ b/lib/kafo/kafo_configure.rb
@@ -22,6 +22,7 @@ require 'kafo/wizard'
 require 'kafo/system_checker'
 require 'kafo/puppet_command'
 require 'kafo/puppet_log_parser'
+require 'kafo/puppet_report'
 require 'kafo/progress_bar'
 require 'kafo/hooking'
 require 'kafo/exit_handler'
@@ -472,7 +473,13 @@ module Kafo
       @progress_bar.close if @progress_bar
       logger.info "Puppet has finished, bye!"
 
+      last_report = execution_env.reports.last
+      if last_report
+        report = PuppetReport.load_report_file(report)
+      end
+
       self.class.exit(exit_code) do
+        # TODO: make report available in the post context
         self.class.hooking.execute(:post)
       end
     end
--- a/lib/kafo/puppet_report.rb
+++ b/lib/kafo/puppet_report.rb
@@ -1,0 +1,34 @@
+module Kafo
+  # An abstraction over the Puppet report format
+  #
+  # @see https://puppet.com/docs/puppet/6.18/format_report.html
+  class PuppetReport
+    attr_reader :report
+
+    def self.load_report_file(path)
+      raise ArgumentError, 'No path given' unless path || path.empty?
+      raise ArgumentError, "#{path} is not a readable file" unless File.file?(path) && File.readable?(path)
+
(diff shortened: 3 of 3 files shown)


ekohl:
@ehelms I talked about this today and revisited it. I've added tests to it (including acceptance tests) so I'm fairly certain this works.

A follow up step would be to parse the report for failed resources and display that logically. I wonder if this should be part of Kafo itself or part of a hook. Any thoughts on that?

ekohl:
It looks like `test/fixtures/hooks/post/20-puppet-report.rb` is ran twice, which surprises me. Other hooks only appear to run once.

test/fixtures/manifests/failure.pp
```diff
@@ -0,0 +1,37 @@
...
+#
+# @param fail
+#   The ways to trigger a failure
+class testing (
+  Array[Enum['exec']] $fail = [],
+) {
+  $fail.each |$failure| {
+    case $failure {
+      'exec': {
+        $command = '/tmp/failing-command'
```
ekohl, line 11: I don't like this hardcoded path, but didn't find an easy way yet. Not that I looked very hard.

lib/kafo/execution_environment.rb
```diff
@@ -19,6 +19,15 @@ def directory
...
     end
 
+    def reportdir
+      @reportdir ||= File.join(directory, 'reports')
+    end
+
+    def reports
+      # Reports are stored in $reportdir/$certname/$report
+      Dir.glob(File.join(reportdir, '*', '*.*')).sort_by { |path| File.mtime(path) }
+    end
```
ehelms, line 29: Do reports get cleaned up automatically over time?

lib/kafo/hook_context.rb
```diff
@@ -196,5 +196,13 @@ def scenario_data
     def exit_code
       self.kafo.exit_code
     end
+
+    # Return the Puppet report, if any.
+    # Only available after Puppet actual ran.
+    #
+    # @return [Optional[Kafo::PuppetReport]]
+    def puppet_report
```
ehelms, line 204: If you try to call this in any of the pre-hooks, I assume it will just return `nil` to indicate there is no report yet? Whereas in any post hook this would be populated by the most recent report?

ehelms:
> A follow up step would be to parse the report for failed resources and display that logically. I wonder if this should be part of Kafo itself or part of a hook. Any thoughts on that?

If we can forsee a reason that the reporting would be contextual to the scenario or configuration of a given scenario then hooks would be a better place. No examples come to mind, so I lean towards Kafo. If we wanted to hedge our bets, we could do Kafo + a way to override that output.

lib/kafo/hook_context.rb
ekohl: That is indeed what I would think, though I didn't test it explicitly.

lib/kafo/execution_environment.rb
ekohl: It is within the execution environment, which is all in a temporary directory that gets cleaned up at the exit of the program:
https://github.com/theforeman/kafo/blob/78f565eeaf6dfb405a9dafcfc52cde4d82e9f5a0/lib/kafo/kafo_configure.rb#L486-L487

ekohl:
I can imagine we want different output handling for `foreman-proxy-certs-generate` than for `foreman-installer`. Perhaps we should have a function in Kafo that you can easily call in a hook. I'll think a bit more about it.

ehelms:
Would this report allow us to make better use of `--noop` and present the what will happen to users?

ekohl:
Oh, excellent point. I didn't think about that. It would, but it might be very long (especially on a fresh installation) and not completely correct.

Though I wonder if that's a path we should open. If we provide that it implies users should care, but should they?

There are some edge cases. For example,with both ISC DHCP and ISD BIND our management is poor. With both the installer is very limited in what it can configure and users often want more. The installer may revert that. On the other hand, there are better ways of solving that.

ISC DHCP is EOL. While still supported on EL8 & EL9, we'll need something else in EL10.

ISC BIND is production quality software that will remain, but we also have many other ways. We only communicate with it using RFC 2136 (via `nsupdate`) so we can recommend separating it for production setups. We have providers for PowerDNS and Infoblox. I think we can also integrate with Red Hat IdM, or implement that if needed.

I'd rather move to recommending that and narrowing the scope of the installer.

ehelms:
> Though I wonder if that's a path we should open. If we provide that it implies users should care, but should they?

I am thinking about this from the post initial installation stand-point. One way to reduce being scared to run the installer is by providing a way to understand what the installer will do if you run it. I was looking at `--noop` as a way for users to get an understanding of what will happen and help them to make choices about doing maintenance. For example, If that plus some clean output helped them understand what services would be restarted they could better plan.

I think the points around the edge cases is a good discussion point regardless, narrowing scope is helpful to us and users.

ekohl:
I've thought about it and I'd like to keep the exact reporting out of Kafo for now. This allows us to iterate quicker. In the future we can provide a better API. One example we can take here the failure message we have: https://github.com/theforeman/foreman-installer/blob/develop/hooks/post/99-post_install_message.rb

Now that the tests are green I'd like to merge & release this.

ehelms (approved):

PR #291, opened 2020-10-14
Allow configuring verbose in config file

ehelms:
To allow a scenario to configure verbose mode by default, verbose has to be allowed as a parameter in the config and the code handling the input from the user has to recognize when to overwrite this from the user and when to avoid it due to potential default answers.

### DIFF
--- a/lib/kafo/configuration.rb
+++ b/lib/kafo/configuration.rb
@@ -32,6 +32,7 @@ module Kafo
         :custom               => {},
         :facts                => {},
         :low_priority_modules => [],
+        :verbose              => false,
         :verbose_log_level    => 'notice',
         :skip_puppet_version_check => false,
         :parser_cache_path    => nil,
--- a/lib/kafo/kafo_configure.rb
+++ b/lib/kafo/kafo_configure.rb
@@ -331,7 +331,8 @@ module Kafo
                  :default => false
       app_option ['--skip-puppet-version-check'], :flag, 'Skip check for compatible Puppet versions',
                  :default => false, :advanced => true
-      app_option ['-v', '--verbose'], :flag, 'Display log on STDOUT instead of progressbar'
+      app_option ['-v', '--[no-]verbose'], :flag, 'Display log on STDOUT instead of progressbar',
+                 :default => config.app[:verbose]
       app_option ['-l', '--verbose-log-level'], 'LEVEL', 'Log level for verbose mode output',
                  :default => 'notice'
       app_option ['-S', '--scenario'], 'SCENARIO', 'Use installation scenario'
--- a/test/acceptance/kafo_configure_test.rb
+++ b/test/acceptance/kafo_configure_test.rb
@@ -58,6 +58,13 @@ module Kafo
         _(code.exitstatus).must_equal 20, err
         _(File.exist?("#{INSTALLER_HOME}/testing")).must_equal false
       end
+
+      it 'must default verbose to false' do
+        code, _, err = run_command '../bin/kafo-configure'
+
+        _(code).must_equal 0, err
+        _(YAML.load_file(KAFO_CONFIG)[:verbose]).must_equal false
+      end
     end
 
     describe '--noop' do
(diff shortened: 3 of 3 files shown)


lib/kafo/kafo_configure.rb
```diff
@@ -375,7 +375,12 @@ def parse_app_arguments
       self.class.app_options.each do |option|
         name                    = option.attribute_name
         value                   = send(option.flag? ? "#{name}?" : name)
-        config.app[name.to_sym] = value.nil? ? option.default_value : value
+
+        if value.nil? && !option.default_value.nil?
```
wbclark, line 379:
```suggestion
        if value.nil?
```
This would be closer to the previous behavior, yes?

If we do in fact want to add some new additional condition `&& !option.default_value.nil?`, shouldn't we also have some handling for when `value.nil? && option.default_value.nil?` such as a warning that the default value is undefined or an error that says this code path shouldn't have been traversed.

ehelms: I have updated this logic to hopefully make it clearer what is different here now and some test cases

lib/kafo/configuration.rb
```diff
@@ -30,7 +30,8 @@ class Configuration
         :facts                => {},
         :low_priority_modules => [],
         :verbose_log_level    => 'info',
-        :skip_puppet_version_check => false
+        :skip_puppet_version_check => false,
+        :verbose              => false,
```
ekohl, line 34: Nit: I think it's better to keep this together with `verbose_log_level`

ekohl (commented):
I don't see any test where `verbose` is true in the config file, but would like to see that as well.

When `verbose` is in the config file, does is there also a `--no-verbose` (or `--quiet`) to disable it again?

lib/kafo/kafo_configure.rb
```diff
@@ -375,7 +375,14 @@ def parse_app_arguments
       self.class.app_options.each do |option|
         name                    = option.attribute_name
         value                   = send(option.flag? ? "#{name}?" : name)
-        config.app[name.to_sym] = value.nil? ? option.default_value : value
+
+        if config.app[name.to_sym]
```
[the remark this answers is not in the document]
ehelms: Only options listed here -- https://github.com/theforeman/kafo/pull/291/files#diff-681434a57726b7033941343071325fac71bcf3c7e02629d6678b082d4ea24e30R13 are affected, will see if I can get tests that show this

test/acceptance/kafo_configure_test.rb
```diff
@@ -68,6 +76,49 @@ module Kafo
...
+      end
+
+      it 'must respect --no-verbose' do
+        code, stdout, err = run_command '../bin/kafo-configure --no-verbose'
+        updated_config = YAML.load_file(@config_file)
+
+        _(code).must_equal 0, err
+        _(stdout).wont_include "INFO "
+        _(updated_config[:verbose]).must_equal false
+      end
```
ekohl, line 119: The `describe` block above suggests it is in the config file but the code looks like it isn't. This should be a new `describe` block, right?

lib/kafo/kafo_configure.rb
```diff
@@ -307,7 +307,7 @@ def set_app_options
                             :default => false
       self.class.app_option ['-s', '--skip-checks-i-know-better'], :flag, 'Skip all system checks', :default => false
       self.class.app_option ['--skip-puppet-version-check'], :flag, 'Skip check for compatible Puppet versions', :default => false
-      self.class.app_option ['-v', '--verbose'], :flag, 'Display log on STDOUT instead of progressbar'
+      self.class.app_option ['-v', '--[no-]verbose'], :flag, 'Display log on STDOUT instead of progressbar'
```
ekohl, line 310: :+1:

ehelms:
This fix here is needed and I split it out to it's own change -- https://github.com/theforeman/kafo/pull/292

lib/kafo/kafo_configure.rb
```diff
@@ -143,6 +143,8 @@ def initialize(*args)
       # so we limit parsing only to app config options (because of --help and later defined params)
       parse clamp_app_arguments
       parse_app_arguments # set values from ARGS to config.app
+      config.configure_application
```
ekohl, line 146: Why is it needed to save the configuration at this point?

ehelms: If a user passes in arguments from the command line, we need to ensure that those values, for any options that are marked for persistence in the config file, get saved to the config file.

ekohl: It feels wrong to save the config before running the checks etc. I tried to look in the docs but it doesn't say anything about it. Perhaps it should only be saved after `pre_commit`?

My reasoning is that the answers aren't stored until then either and it's still possible to abort the run without making modifications.

ehelms: I think that's fair

ehelms:
Rebased on master, this is now just adding verbose to the saved config and defaulting it to false to match currently how it works.

wbclark (approved):
This works nicely with #292 and the tests seem to cover all bases. It looks good to me.

PR #292, opened 2020-10-19
Only save configuration values defined to be persisted

ehelms:
The entire application config can be persisted to disk but not every
item is meant to be saved to disk as they are designed to be
options the user passes during runtime and not enabled every
single run. This ensures only configuration defined in the
DEFAULTS hash is persisted to disk.

This also prevents the false assumption that changing a saved value
in the config will affect the next run which is not always true for
every value.

### DIFF
--- a/lib/kafo/configuration.rb
+++ b/lib/kafo/configuration.rb
@@ -15,6 +15,8 @@ module Kafo
         :description          => '',
         :enabled              => true,
         :log_dir              => '/var/log/kafo',
+        :log_owner            => nil,
+        :log_group            => nil,
         :store_dir            => '',
         :log_name             => 'configuration.log',
         :log_level            => 'notice',
@@ -26,11 +28,19 @@ module Kafo
         :colors               => Kafo::ColorScheme.colors_possible?,
         :color_of_background  => :dark,
         :hook_dirs            => [],
+        :check_dirs           => nil,
         :custom               => {},
         :facts                => {},
         :low_priority_modules => [],
         :verbose_log_level    => 'notice',
-        :skip_puppet_version_check => false
+        :skip_puppet_version_check => false,
+        :parser_cache_path    => nil,
+        :ignore_undocumented  => nil,
+        :order                => nil,
+        :hiera_config         => nil,
+        :kafo_modules_dir     => nil,
+        :config_header_file   => nil,
+        :dont_save_answers    => nil,
     }
 
     def self.get_scenario_id(filename)
@@ -57,11 +67,16 @@ module Kafo
 
     def save_configuration(configuration)
       return true unless @persist
+
+      trimmed = configuration.reject do |key, value|
+        !DEFAULT.key?(key) || value.nil?
+      end
+
       begin
         FileUtils.touch @config_file
         File.chmod 0600, @config_file
         File.open(@config_file, 'w') do |file|
-          file.write(format(YAML.dump(configuration.sort.to_h)))
+          file.write(format(YAML.dump(trimmed.sort.to_h)))
         end
       rescue Errno::EACCES
         puts "Insufficient permissions to write to #{@config_file}, can not continue"
--- a/lib/kafo/kafo_configure.rb
+++ b/lib/kafo/kafo_configure.rb
@@ -198,6 +198,7 @@ module Kafo
 
       self.class.hooking.execute(:pre_commit)
       unless dont_save_answers? || noop?
+        config.configure_application
         store_params
         self.class.scenario_manager.link_last_scenario(self.class.config_file) if self.class.scenario_manager.configured?
       end
@@ -312,13 +313,13 @@ module Kafo
 
     def set_app_options
       app_option ['--[no-]colors'], :flag, 'Use color output on STDOUT',
-                 :default => !!config.app[:colors], :advanced => true
+                 :default => config.app[:colors], :advanced => true
       app_option ['--color-of-background'], 'COLOR', 'Your terminal background is :bright or :dark',
                  :default => config.app[:color_of_background], :advanced => true
       app_option ['--dont-save-answers'], :flag, "Skip saving answers to '#{self.class.config.answer_file}'?",
-                 :default => !!config.app[:dont_save_answers], :advanced => true
+                 :default => config.app[:dont_save_answers], :advanced => true
       app_option '--ignore-undocumented', :flag, 'Ignore inconsistent parameter documentation',
-                 :default => false, :advanced => true
+                 :default => config.app[:ignore_undocumented], :advanced => true
       app_option ['-i', '--interactive'], :flag, 'Run in interactive mode'
       app_option '--log-level', 'LEVEL', 'Log level for log file output',
                  :default => config.app[:log_level], :advanced => true
--- a/test/acceptance/kafo_configure_test.rb
+++ b/test/acceptance/kafo_configure_test.rb
@@ -173,5 +173,57 @@ module Kafo
         _(File.exist?("#{INSTALLER_HOME}/testing")).must_equal true
       end
     end
+
+    describe 'the configuration file' do
+      before do
+        @saved_values = Kafo::Configuration::DEFAULT.reject do |key, value|
+          value.nil?
+        end
+      end
+
+      it 'must only save defined values' do
+        code, _, err = run_command '../bin/kafo-configure'
+
+        _(code).must_equal 0, err
+        _(YAML.load_file(KAFO_CONFIG).keys.sort).must_equal @saved_values.keys.sort
+      end
+
+      it 'must not save command line options when --noop' do
+        config = YAML.load_file(KAFO_CONFIG)
+        config[:color_of_background] = "bright"
+        File.open(KAFO_CONFIG, 'w') do |file|
+          file.write(config.to_yaml)
+        end
+
+        code, _, err = run_command '../bin/kafo-configure --noop --color-of-background dark'
+
+        _(code).must_equal 0, err
+        _(YAML.load_file(KAFO_CONFIG).keys.sort).must_equal @saved_values.keys.sort
+        _(YAML.load_file(KAFO_CONFIG).keys).wont_include 'noop'
+        _(YAML.load_file(KAFO_CONFIG)[:color_of_background]).must_equal "bright"
+      end
+
+      it 'must not save non-persisted command line options' do
+        code, _, err = run_command '../bin/kafo-configure --profile'
+
+        _(code).must_equal 0, err
+        _(YAML.load_file(KAFO_CONFIG).keys.sort).must_equal @saved_values.keys.sort
+        _(YAML.load_file(KAFO_CONFIG).keys).wont_include 'profile'
+      end
+
+      it 'must save persisted command line options' do
+        config = YAML.load_file(KAFO_CONFIG)
+        config[:color_of_background] = "bright"
+        File.open(KAFO_CONFIG, 'w') do |file|
+          file.write(config.to_yaml)
+        end
+
+        code, _, err = run_command '../bin/kafo-configure --color-of-background dark'
+
+        _(code).must_equal 0, err
+        _(YAML.load_file(KAFO_CONFIG).keys.sort).must_equal @saved_values.keys.sort
+        _(YAML.load_file(KAFO_CONFIG)[:color_of_background]).must_equal "dark"
+      end
+    end
   end
 end


wbclark (commented):
@ehelms can you provide an example or two of application options that aren't defined in `DEFAULT` and would therefore be affected by this change? I'm looking for context to understand why we didn't do this before and why we are making the change now.

Is this change in response to a specific bug or issue, and if so can we attach an issue # ?

ehelms:
Any options defined in a hook are not persisted. Then you have options such as `--list-scenarios` or `--noop` or `---dont-save-answers` that a user may want to trigger on a single run but not have saved for every single run.

Hooks can call out to save the current configuration to disk (https://github.com/theforeman/foreman-installer/blob/develop/hooks/pre_commit/13-tuning.rb#L60) which writes all current app configuration to the config file. However, not every value there is respected. For example, noop can get written to the config file but on the next run if a user does not supply `--noop` then, even though it is `noop: true` in the config file, the installer is not run in noop mode *as expected*. So this prevents the user having a false sense of persisted configurations.

ekohl (commented):
This would be a breaking change.

How would a hook ensure an app option is persisted? At least one hook's documentation suggests it relies on it:

https://github.com/theforeman/foreman-installer/blob/814f75071305b9ca70cd20ba62423496c67777f4/hooks/boot/09-version_locking.rb#L10-L11

ehelms:
Prior to this change, custom hook based options could be *saved* to the config file, but they were not respected. If we look at package locking for example, that makes use of the custom facts to track the state between runs (https://github.com/theforeman/foreman-installer/blob/814f75071305b9ca70cd20ba62423496c67777f4/hooks/pre_commit/09-version_locking.rb#L24)

And that is part of my point with this PR, there is a false assumption that all values that get stored to the config file persist between runs because some hooks are calling `configure_application` which dumps the entire config to the config file. But that, without this bug fix, does not guarantee everything in there is persistent. So from all my testing, no existing hook would break because it did not work.

lib/kafo/configuration.rb
```diff
@@ -57,10 +59,13 @@ def initialize(file, persist = true)
 
     def save_configuration(configuration)
       return true unless @persist
+
+      trimmed = configuration.reject { |key, value| !DEFAULT.key?(key) }
```
ekohl, line 63: Isn't this equal, but more efficient?
```suggestion
      trimmed = configuration.slice(*DEFAULT.keys)
```

However, you may want to consider skipping values that are `nil`.

lib/kafo/configuration.rb
```diff
@@ -30,7 +30,9 @@ class Configuration
         :facts                => {},
         :low_priority_modules => [],
         :verbose_log_level    => 'info',
-        :skip_puppet_version_check => false
+        :skip_puppet_version_check => false,
+        :parser_cache_path    => nil,
+        :ignore_undocumented  => nil,
```
ekohl, line 35: Just comparing to https://github.com/theforeman/foreman-installer/blob/develop/config/foreman.yaml I think you're missing:
* `hiera_config`
* `log_owner`
* `log_group`
* `config_header_file`
* `dont_save_answers`
* `kafo_modules_dir`
* `order`

There may be more.

ekohl (changes requested):
This is a point that always confused me and thought was ugly.

>  If we look at package locking for example, that makes use of the custom facts to track the state between runs (https://github.com/theforeman/foreman-installer/blob/814f75071305b9ca70cd20ba62423496c67777f4/hooks/pre_commit/09-version_locking.rb#L24)

custom config != custom facts. Otherwise point taken.

I do wonder if `app_option` should have a way to indicate it should be stored. That way we don't need custom configs for that. Not a blocker for this, but probably a nice follow up.

ehelms:
The current list of configuration options that get written to disk are:

```
:name: Foreman
:description: Default installation of Foreman
:enabled: true
:log_dir: "./_build/"
:store_dir: ''
:log_name: foreman.log
:log_level: :debug
:no_prefix: false
:mapping: {}
:answer_file: "./config/foreman-answers.yaml"
:installer_dir: "."
:module_dirs: "./_build/modules"
:colors: true
:color_of_background: :dark
:hook_dirs: []
:custom: {}
:facts: {}
:low_priority_modules:
- foreman_proxy_plugin
- foreman_compute
- foreman_plugin
:verbose_log_level: info
:skip_puppet_version_check: false
:parser_cache_path: "./_build/parser_cache/foreman.yaml"
:ignore_undocumented: false
:log_owner: foreman
:reset_data: false
:detailed_exitcodes: false
:clear_pulp_content: false
:dont_save_answers: false
:interactive:
:noop: false
:profile: false
:skip_checks_i_know_better: false
:verbose: true
:scenario:
:disable_scenario:
:enable_scenario:
:list_scenarios:
:force:
:compare_scenarios:
:migrations_only:
:parser_cache:
```

The current set that would get persisted by this change are:

```
:name: Foreman
:description: Default installation of Foreman
:enabled: true
:log_dir: "./_build/"
:store_dir: ''
:log_name: foreman.log
:log_level: :debug
:no_prefix: false
:mapping: {}
:answer_file: "./config/foreman-answers.yaml"
:installer_dir: "."
:module_dirs: "./_build/modules"
:colors: true
:color_of_background: :dark
:hook_dirs: []
:custom: {}
:facts: {}
:low_priority_modules:
- foreman_proxy_plugin
- foreman_compute
- foreman_plugin
:verbose_log_level: info
:skip_puppet_version_check: false
:parser_cache_path: "./_build/parser_cache/foreman.yaml"
:ignore_undocumented: false
```

The set that would not get persisted are:

```
:reset_data: false
:detailed_exitcodes: false
:clear_pulp_content: false
:dont_save_answers: false
:interactive:
:noop: false
:profile: false
:skip_checks_i_know_better: false
:verbose: true
:scenario:
:disable_scenario:
:enable_scenario:
:list_scenarios:
:force:
:compare_scenarios:
:migrations_only:
:parser_cache:
```
Of this list, I do not think we want to store any of these between runs.

For some of the ones you mentioned like `log_owner` or `log_group` my testing indicates even if you set those today they are not respected (and why they do not show up in either list above) and I believe that is due to a bug in that if its not listed in the DEFAULT hash it does not get respected as an option. I do not know how long that has been true, and we should re-evaluate and fix that for parameters that fall into that category. However, this change I believe does not change that behavior for those parameters and they should be treated independently.

ekohl:
At least `kafo_modules_dir` is used on Debian in production builds. We also set `hiera_config`, `parser_cache_path` and `order. You do not mention those.

ehelms:
Added the ones I was missing from some repeated runs to the default, and added a couple of tests. The default list of saved items now comes out for a standard Foreman config as:

```
---
:name: Foreman
:description: Default installation of Foreman
:enabled: true
:log_dir: "./_build/"
:store_dir: ''
:log_name: foreman.log
:log_level: :debug
:no_prefix: false
:mapping: {}
:answer_file: "./config/foreman-answers.yaml"
:installer_dir: "."
:module_dirs: "./_build/modules"
:colors: true
:color_of_background: :dark
:hook_dirs: []
:check_dirs:
:custom: {}
:facts: {}
:low_priority_modules:
- foreman_proxy_plugin
- foreman_compute
- foreman_plugin
:verbose_log_level: info
:skip_puppet_version_check: false
:parser_cache_path: "./_build/parser_cache/foreman.yaml"
:ignore_undocumented: false
:order:
- foreman
- foreman_proxy
- puppet
:hiera_config: "./config/foreman-hiera.yaml"
:kafo_modules_dir:
:config_header_file:
```

lib/kafo/configuration.rb
ehelms: Slice for hash was added in Ruby 2.5, we still test against Ruby 2.4 so we have to stick with the current implementation.

ekohl: There must be some joke about sliced bread in there.

lib/kafo/configuration.rb
```diff
@@ -57,10 +64,13 @@ def initialize(file, persist = true)
 
     def save_configuration(configuration)
       return true unless @persist
+
+      trimmed = configuration.reject { |key, value| !DEFAULT.key?(key) }
```
ekohl, line 68: Does it make sense to ignore nil values?
```suggestion
      trimmed = configuration.reject { |key, value| !DEFAULT.key?(key) || value.nil? }
```

ekohl (commented):
> For some of the ones you mentioned like `log_owner` or `log_group` my testing indicates even if you set those today they are not respected

I can't reproduce that:

```console
# ls -l /var/log/foreman-installer/foreman.log 
-rw-r--r--. 1 root root 25155 Oct 21 15:14 /var/log/foreman-installer/foreman.log
# tail -n 2 /etc/foreman-installer/scenarios.d/foreman.yaml 
:log_owner: ekohl
:log_group: ekohl
# foreman-installer --help
Could not get default values, check log file at /var/log/foreman-installer/foreman.log for more information
# ls -l /var/log/foreman-installer/foreman.log 
-rw-r--r--. 1 ekohl ekohl 25155 Oct 21 15:16 /var/log/foreman-installer/foreman.log
```

It still feels weird to ignore parameters documented in the example:
https://github.com/theforeman/kafo/blob/69880ef78ffff80f9d2c0c0bd93fe8f63890d0fd/config/kafo.yaml.example#L39-L44

For `log_owner` / `log_group` I don't foresee any problems with adding them, but I wonder about `dont_save_answers`. Consider the following scenario:
* A config is migrated (AFAIK the Satellite migration tool does this)
* Some hook stores the config (by calling `kafo.config.configure_application`)

Now the scenario will never store the answers again. Worse, the user will not immediately notice. Only on the next run they may (or may not) notice that the option was reset.

lib/kafo/configuration.rb
ehelms: If we ignore nil, we should ignore empty arrays, empty hashes, empty strings. I'd rather just dump the configuration and it's current set of values for values that we persist to the file.

ehelms:
I have added log_owner, log_group, and dont_save_answers.

ehelms:
> A config is migrated (AFAIK the Satellite migration tool does this)
Some hook stores the config (by calling kafo.config.configure_application)
Now the scenario will never store the answers again. Worse, the user will not immediately notice. Only on the next run they may (or may not) notice that the option was reset.

I think  for a user it seems strange to persist this, however, we do persist it already as we have hooks that do cause the saving. The value is nil so by default it does not have any affects. If a user does go and use this parameter, then it will persist between runs but that is currently how this is designed to work anyway. So I do not think we are adding additional concerns to what already exists.

ekohl:
> I think for a user it seems strange to persist this, however, we do persist it already as we have hooks that do cause the saving. The value is nil so by default it does not have any affects. If a user does go and use this parameter, then it will persist between runs but that is currently how this is designed to work anyway. So I do not think we are adding additional concerns to what already exists.

In itself, it's fine in this PR. However, it may be a concern in https://github.com/theforeman/kafo/pull/291 when the app option suddenly does respect the default.

I think this PR is almost good to merge. After dinner I'll try to test it out in the latest updated version. We can take the discussion to the other PR.

ekohl (approved):
I ran this on a nightly, copied `/etc/foreman-installer` to `/etc/foreman-installer.old`, installed http://koji.katello.org/koji/taskinfo?taskID=377571 and ran it again.

```console
# foreman-installer --verbose
2020-10-22 15:08:29 [NOTICE] [pre_migrations] Executing hooks in group pre_migrations
2020-10-22 15:08:29 [NOTICE] [pre_migrations] All hooks in group pre_migrations finished
2020-10-22 15:08:29 [NOTICE] [boot] Executing hooks in group boot
2020-10-22 15:08:29 [ERROR ] [root] Command foreman-maintain packages -h not found
2020-10-22 15:08:29 [NOTICE] [boot] All hooks in group boot finished
2020-10-22 15:08:29 [NOTICE] [init] Executing hooks in group init
2020-10-22 15:08:29 [NOTICE] [init] All hooks in group init finished
2020-10-22 15:08:29 [NOTICE] [root] Loading default values from puppet modules...
2020-10-22 15:08:33 [NOTICE] [root] ... finished
2020-10-22 15:08:33 [NOTICE] [pre_values] Executing hooks in group pre_values
2020-10-22 15:08:33 [NOTICE] [pre_values] All hooks in group pre_values finished
2020-10-22 15:08:34 [NOTICE] [pre_validations] Executing hooks in group pre_validations
2020-10-22 15:08:35 [NOTICE] [pre_validations] All hooks in group pre_validations finished
2020-10-22 15:08:35 [NOTICE] [root] Running validation checks
2020-10-22 15:08:35 [NOTICE] [pre_commit] Executing hooks in group pre_commit
2020-10-22 15:08:35 [NOTICE] [pre_commit] All hooks in group pre_commit finished
2020-10-22 15:08:35 [NOTICE] [pre] Executing hooks in group pre
2020-10-22 15:08:39 [NOTICE] [pre] All hooks in group pre finished
2020-10-22 15:08:39 [NOTICE] [configure] Starting system configuration
2020-10-22 15:10:56 [NOTICE] [configure] Puppet has finished, bye!
2020-10-22 15:10:56 [NOTICE] [post] Executing hooks in group post
Executing: foreman-rake upgrade:run
foreman-rake upgrade:run finished successfully!
  Success!
  * Foreman is running at https://centos7-foreman-nightly.wisse.example.com
  * Foreman Proxy is running at https://centos7-foreman-nightly.wisse.example.com:8443

  The full log is at /var/log/foreman-installer/foreman.log
2020-10-22 15:11:02 [NOTICE] [post] All hooks in group post finished
```

This produces the following diff:

```diff
diff -Nutr foreman-installer.old/scenarios.d/foreman.yaml foreman-installer/scenarios.d/foreman.yaml
--- foreman-installer.old/scenarios.d/foreman.yaml	2020-10-22 15:07:23.846563614 +0000
+++ foreman-installer/scenarios.d/foreman.yaml	2020-10-22 15:08:29.445015191 +0000
@@ -3,6 +3,8 @@
 :description: Default installation of Foreman
 :enabled: true
 :log_dir: "/var/log/foreman-installer"
+:log_owner: 
+:log_group: 
 :store_dir: ''
 :log_name: foreman.log
 :log_level: :debug
@@ -14,17 +16,23 @@
 :colors: true
 :color_of_background: :dark
 :hook_dirs: []
+:check_dirs: 
 :custom: {}
 :facts: {}
 :low_priority_modules:
 - foreman_proxy_plugin
 - foreman_compute
 - foreman_plugin
-:verbose_log_level: info
+:verbose_log_level: notice
+:verbose: true
 :skip_puppet_version_check: false
-:hiera_config: "/usr/share/foreman-installer/config/foreman-hiera.yaml"
 :parser_cache_path: "/usr/share/foreman-installer/parser_cache/foreman.yaml"
+:ignore_undocumented: false
 :order:
 - foreman
 - foreman_proxy
 - puppet
+:hiera_config: "/usr/share/foreman-installer/config/foreman-hiera.yaml"
+:kafo_modules_dir: 
+:config_header_file: 
+:dont_save_answers: false
```

I'm still leaning to a logic that roughly comes down to `delete_if { |key, value| DEFAULT[key].nil? && value.nil? }` to keep it smaller but don't strongly insist on it.

For the implementation I would also consider a separate method `trim_configuration_for_storage` and test that as a unit test. Unit tests are typically faster than acceptance tests.

However, in principle this I think this is a good change. I'll leave it up to you to decide if you want to address those implementation details.

ehelms:
I went with the dropping of nil and moved a couple of options to nil instead of false or computed. I think that now meets expectation, @ekohl you did approve but since I made a few modifications I would ask you check those.

https://github.com/theforeman/kafo/pull/292/files#diff-681434a57726b7033941343071325fac71bcf3c7e02629d6678b082d4ea24e30L26-L27

lib/kafo/configuration.rb
```diff
@@ -23,14 +25,20 @@ class Configuration
         :answer_file          => './config/answers.yaml',
         :installer_dir        => '.',
         :module_dirs          => ['./modules'],
-        :colors               => Kafo::ColorScheme.colors_possible?,
```
ekohl, line 26: Don't we still need a default for this? If I read this code correctly, I think we'll end up without colors if you drop this:
https://github.com/theforeman/kafo/blob/9fb119aff2f9827ecd088438aacc613f3c56c8c1/lib/kafo/kafo_configure.rb#L498-L506

For my own understanding, I wrote https://github.com/theforeman/kafo/pull/296. Currently untested, but I think that would be needed.

lib/kafo/configuration.rb
```diff
@@ -23,14 +25,20 @@ class Configuration
...
-        :color_of_background  => :dark,
```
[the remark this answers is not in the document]
ehelms: I can put it back, it seemed somewhat odd for this not to just default unless specified by a user.

lib/kafo/kafo_configure.rb
```diff
@@ -291,7 +291,7 @@ def set_parameters
 
     def set_app_options
       self.class.app_option ['--[no-]colors'], :flag, 'Use color output on STDOUT',
-                            :default => !!config.app[:colors]
+                            :default => Kafo::ColorScheme.colors_possible?
```
ehelms, line 294: @ekohl We can just set the default here passed on the computed value with no need to persist this between runs, that way a user can specify this at runtime if they desire to override or let the app decide if it's possible.

This worked for me with `--no-colors` being passed.

ekohl: This removes the option for a scenario to always disable colors, which was previously possible.

ehelms: I put it back the way it has been. Let's just leave colors as an option alone the way it worked before.

ehelms:
Rebased on top of the AppOption change

lib/kafo/configuration.rb
```diff
@@ -15,22 +15,32 @@ class Configuration
...
         :log_level            => 'notice',
         :no_prefix            => false,
         :mapping              => {},
         :answer_file          => './config/answers.yaml',
         :installer_dir        => '.',
-        :module_dirs          => ['./modules'],
-        :colors               => Kafo::ColorScheme.colors_possible?,
         :color_of_background  => :dark,
+        :colors               => Kafo::ColorScheme.colors_possible?,
+        :module_dirs          => ['./modules'],
```
ekohl, line 29: Is there a reason to reorder these?

test/acceptance/kafo_configure_test.rb
```diff
@@ -5,6 +5,7 @@ module Kafo
     before do
       generate_installer
       add_manifest
+      @config_file = "#{INSTALLER_HOME}/config/installer-scenarios.d/default.yaml"
```
ekohl, line 8: I think there's already a constant `KAFO_CONFIG` for this:
https://github.com/theforeman/kafo/blob/dd1f1f73548f1e2a37e1ffc9fdf3f6d977f2d0bd/test/acceptance/test_helper.rb#L9

test/acceptance/kafo_configure_test.rb
```diff
@@ -173,5 +174,27 @@ module Kafo
...
+
+        _(code).must_equal 0, err
+        _(YAML.load_file(@config_file).keys.sort).must_equal @saved_values.keys.sort
+      end
+
+      it 'must not save non persisted command line options' do
+        code, _, err = run_command '../bin/kafo-configure --noop'
+
+        _(code).must_equal 0, err
+        _(YAML.load_file(@config_file).keys.sort).must_equal @saved_values.keys.sort
```
ekohl, line 196: I think you should also check that `noop` is still `false`, otherwise the test doesn't really say anything compared to the other test.

ekohl (commented):
Could you add a test that verifies that new app values are still being stored? For example, changing the color or color background by passing it in via a command line option. With that, I think I'd feel safe to merge it.

test/acceptance/kafo_configure_test.rb
ehelms: I added a check that verifies `noop` is nil in the config file.

test/acceptance/kafo_configure_test.rb
```diff
@@ -173,5 +173,41 @@ module Kafo
...
+
+      it 'must not save non persisted command line options' do
+        code, _, err = run_command '../bin/kafo-configure --noop'
+
+        _(code).must_equal 0, err
+        _(YAML.load_file(KAFO_CONFIG).keys.sort).must_equal @saved_values.keys.sort
+        _(YAML.load_file(KAFO_CONFIG).keys).wont_include 'noop'
+      end
+
+      it 'must save persisted command line options' do
```
ehelms, line 199: Additional persisted option test

test/acceptance/kafo_configure_test.rb
```diff
@@ -173,5 +173,41 @@ module Kafo
...
+
+      it 'must only save defined values' do
+        code, _, err = run_command '../bin/kafo-configure'
+
+        _(code).must_equal 0, err
+        _(YAML.load_file(KAFO_CONFIG).keys.sort).must_equal @saved_values.keys.sort
+      end
+
+      it 'must not save non persisted command line options' do
+        code, _, err = run_command '../bin/kafo-configure --noop'
```
ekohl, line 192: Could you also add colors here, similar to the persisted command line options?

ehelms: Check now.. is that what you meant?

test/acceptance/kafo_configure_test.rb
```diff
@@ -173,5 +173,48 @@ module Kafo
...
+        File.open(KAFO_CONFIG, 'w') do |file|
+          file.write(config.to_yaml)
+        end
+
+        code, _, err = run_command '../bin/kafo-configure --noop --no-colors'
+
+        _(code).must_equal 0, err
+        _(YAML.load_file(KAFO_CONFIG).keys.sort).must_equal @saved_values.keys.sort
+        _(YAML.load_file(KAFO_CONFIG).keys).wont_include 'noop'
+        _(YAML.load_file(KAFO_CONFIG)[:colors]).must_equal false
```
ekohl, line 203: This is precisely the kind of test I meant. As a user you might expect that `--noop` doesn't save anything, but here it does. That surprises me.

I do wonder if the autodetection that colors has makes a difference, but using `color_of_background` doesn't have that and can certainly answer the question.

ehelms: Ohh, now I see what you were asking for. I will update the test (note that I am passing --colors so this is exactly as expected).

ehelms: Or are you just pointing to the expectations of `noop` itself? I do not think it works as you are expecting today either and I'm not sure it should. I grant there are multiple levels on which to `interpret` noop, but it's a bit trickier with the options destined for the config file. From testing, if you pass `--noop --ignore-undocumented` today with kafo, the value gets saved to config file. I think this would have to be a separate bug fix.

ehelms: So after taking some consideration from https://github.com/theforeman/kafo/pull/291 and rebasing this on latest, I ended up with a solution that was needed to solve the target of this issue and it in turn addresses your concern here over noop. The tests have been updated to reflect this.

ekohl (approved):
This is indeed what I meant.

ehelms:
This PR is green and back to good to go now after rebasing master on the latest changes.

PR #293, opened 2020-10-20
Fixes #29965 - Support highline 2.x

ekohl:
The biggest change in version 2 is dropping Highline::SystemExtensions in favor of Highline::Terminal.

Another big change is that the global $terminal is no longer set. Instead, it has HighLine.default_instance.

It also changes the highline import to only highline instead of highline/import where possible. This means that not everywhere there is a global agree/ask/choose/say. Since the scenario manager still uses highline/import, for hooks there is no change.

It still keeps highline 1.x support because that's shipped on Debian 10 and Ubuntu 18.04. It also allows for a smoother migration since other gems also use highline.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -4,7 +4,7 @@ load 'tasks/jenkins.rake'
 
 Rake::TestTask.new('test:ruby') do |t|
   t.libs << 'lib' << 'test'
-  t.test_files = FileList['test/**/*_test.rb']
+  t.test_files = FileList['test/**/*_test.rb'] - FileList['test/tmp/**/*_test.rb']
   t.verbose = true
 end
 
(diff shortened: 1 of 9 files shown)


ekohl:
Please review carefully. Because kafo_wizards is not yet converted, it's not yet possible to do a full system test with it.

ekohl:
Looks like the failures are because the progress bar width can't be determined if there is no TTY.

ehelms:
[test kafo]

ekohl:
It's green now, but I should note that I haven't tested this on an actual machine with the installer. Our CI also doesn't cover testing with Highline 1.x

ehelms (approved):
Tested on an install and with highline 2.0.3.

PR #294, opened 2020-10-21
Document HookContext with param and return types

ekohl:
What made me do this was that the examples don't render well (see https://www.rubydoc.info/gems/kafo/5.1.0/Kafo/HookContext). Adding data types makes the documentation a bit more useful.

### DIFF
--- a/lib/kafo/hook_context.rb
+++ b/lib/kafo/hook_context.rb
@@ -3,7 +3,17 @@ require 'kafo/base_context'
 
 module Kafo
   class HookContext < BaseContext
-    attr_reader :kafo, :logger
+    # @return [Kafo::KafoConfigure]
+    attr_reader :kafo
+
+    # some of hooks won't print any message because logger is not yet configured
+    # configuration of logger depends on application configration (log level etc.)
+    #
+    # @return [Kafo::Logger]
+    #
+    # @example
+    #   logger.warn "this combindation of parameters is untested"
+    attr_reader :logger
 
     def self.execute(kafo, logger, &hook)
       new(kafo, logger).instance_eval(&hook)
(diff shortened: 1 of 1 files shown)


lib/kafo/hook_context.rb
```diff
@@ -3,7 +3,17 @@
 
 module Kafo
   class HookContext < BaseContext
-    attr_reader :kafo, :logger
+    # @return [Kafo::KafoConfigure]
+    attr_reader :kafo
+
+    # some of hooks won't print any message because logger is not yet configured
+    # configuration of logger depends on application configration (log level etc.)
```
wbclark, line 10:
```suggestion
    # Hooks in `pre_migrations` and `boot` stages won't print any message because the logger configuration
    # depends on the application configuration, which is not yet available.
```

Might as well improve this a bit while we are here. I'm unsure though if 'available' is the right word here, maybe 'complete' is better.

wbclark:
Not really a blocker, but for my benefit, do you have an example (from another project?) of how these docs are rendered with param and return types?

wbclark (approved):
I'll approve this and leave it up to you if you want to address the above nit before merging.

ekohl:
> Not really a blocker, but for my benefit, do you have an example (from another project?) of how these docs are rendered with param and return types?

[yard](https://yardoc.org/guides/index.html) is used. There is a **lot** more we can do. For example, `@param` also accepts an actual description of each parameter. I thought the names were self documenting enough for now.

lib/kafo/hook_context.rb
ekohl: Initialized might be a better word, but really it shouldn't be used in those stages. I've actually been thinking about how we can actually remove them from the context. Perhaps it should raise an exception.

ehelms: This is a bit confusing as to what it means by won't print any messages. Did this mean in the log file? On the console? When using the progress bar? I know this is true for the progress bar output, but its not true for the log file or console based output that comes with verbose.

ekohl: Given this is only moving the existing documentation I'm going to merge it now. There is plenty of room for more improvement.

PR #295, opened 2020-10-22
Document scenario options

ekohl:
This takes the options from the example scenario and documents them. It also does some grouping of options.

If done right, all code throughout Kafo should use the constants rather than symbols. That way it's easy to see where a scenario option is actuall used throughout the code.


ekohl:
For those unfamiliar with yard: a preview can be rendered using `bundle exec yard && xdg-open doc/Kafo/ScenarioOptions.html`

ehelms:
For me this has benefits and can go in as the starting place to begin using the constants across the code base. If you could rebase @ekohl I'd like to merge this.

lib/kafo/scenario_options.rb
```diff
@@ -0,0 +1,66 @@
+module Kafo
+  # A class containing constants for all scenario options
+  class ScenarioOptions
```
ekohl, line 3: Before I rebase: should this be singular or plural? I think `Kafo::ScenarioOption::NAME` reads better than `Kafo::ScenarioOptions::Name`. Thoughts?

ehelms: Singular.

ehelms:
Could you rebase and address the singular comment above?

ekohl:
Rebased

ehelms (approved):

PR #296, opened 2020-10-23
Determine color usage dynamically

ekohl:
This changes the color usage to be determined dynamically. Prior to this the config was always read. That meant that if the scenario was ever written to disk (using save_configuration), it stored the value of colors_possible? and uses that indefinitely.

After this change, the value of colors will remain nil (determined at runtime) until the user uses --[no-]colors, which is then persisted.

### DIFF
--- a/lib/kafo/kafo_configure.rb
+++ b/lib/kafo/kafo_configure.rb
@@ -125,7 +125,7 @@ module Kafo
       parse_cli_arguments
 
       if !config.app[:verbose]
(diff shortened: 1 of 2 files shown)


lib/kafo/kafo_configure.rb
```diff
@@ -571,6 +571,35 @@ def config_file
       File.join(Dir.pwd, 'config', 'kafo.yaml')
     end
 
+<<<<<<< HEAD
+=======
```
ehelms, line 575: Little diff leftover

PR #297, opened 2020-10-23
Sort configuration when dumping to file

ehelms:


lib/kafo/configuration.rb
```diff
@@ -60,7 +60,9 @@ def save_configuration(configuration)
       begin
         FileUtils.touch @config_file
         File.chmod 0600, @config_file
-        File.open(@config_file, 'w') { |file| file.write(format(YAML.dump(configuration))) }
+        File.open(@config_file, 'w') do |file|
+          file.write(format(YAML.dump(configuration.sort.to_h)))
```
wbclark, line 64: What is the benefit of calling `to_h` here when it wasn't used before?

ehelms: Without the `to_h` this will write itself as an array. The sort turns the hash into a sorted array, the to_h turns it back into a hash.

wbclark: OK, I tested that and I see it's indeed the case. I read in some older tutorial that `sort` doesn't handle hashes (hence my confusion about the type of `configuration` to begin with) with symbol keys well but it's working just fine for me at the ruby console.

Any reason that you left this as a draft?

ehelms: Originally, I left it as a draft just do the amount of other PRs touching this area and wishing to get them in first. But I will switch out and deal with rebases.

wbclark (approved):

tbrisker:
This broke puppet 4 on ruby 2.0, though tbh i don't know why we're still testing that: https://ci.theforeman.org/job/test_kafo_master/39/

ehelms:
I didn't realize we still had that job, and it's out of sync with the pr-test matrix (https://ci.theforeman.org/job/kafo-pr-test/). I will tackle that.

ehelms:
Here ya go -- https://github.com/theforeman/foreman-infra/pull/1476

PR #298, opened 2020-10-24
Move class methods in KafoConfigure together

ehelms:


ehelms:
[test kafo]

wbclark (commented):
@ehelms could you provide a brief explanation of how this supports #299 ?

ehelms:
Overall, this is just a clean up re-factoring that makes following or seeing where the code lives easier. It only supports #299 in the case that due to the changes around options in it, this re-organization of the code makes that easier to follow and easier to clean up.

wbclark (approved):
Thanks @ehelms . I also wondered about the change from `def self.method_name` to `def method_name` but then I saw we already use `class << self` above.

This looks good to me.

PR #299, opened 2020-10-24
App option

ehelms:
Includes https://github.com/theforeman/kafo/pull/298 

There are two commits, one to add AppOption, and one to add advanced. The advanced is tightly coupled to the introduction to AppOption and thus I have kept them together but as separate commits.

### DIFF
--- a/lib/kafo/kafo_configure.rb
+++ b/lib/kafo/kafo_configure.rb
@@ -28,14 +28,17 @@ require 'kafo/exit_handler'
 require 'kafo/scenario_manager'
 require 'kafo/execution_environment'
 require 'kafo/logging'
+require 'kafo/app_option/declaration'
 
 module Kafo
   class KafoConfigure < Clamp::Command
     include StringHelper
 
     class << self
+      include AppOption::Declaration
+
       attr_accessor :config, :root_dir, :config_file, :gem_root,
-                    :module_dirs, :kafo_modules_dir, :verbose, :app_options, :logger,
+                    :module_dirs, :kafo_modules_dir, :verbose, :logger,
                     :check_dirs, :exit_handler, :scenario_manager, :store
       attr_writer :hooking
 
@@ -66,12 +69,6 @@ module Kafo
         builder_class = kafo.full_help? ? HelpBuilders::Advanced : HelpBuilders::Basic
         args.push builder_class.new(kafo.params)
         super(*args)
-      end
-
-      def app_option(*args, &block)
-        self.app_options ||= []
-        self.app_options.push self.option(*args, &block)
-        self.app_options.last
       end
 
       def use_colors?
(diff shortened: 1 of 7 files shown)


lib/kafo/kafo_configure.rb
```diff
@@ -68,12 +71,6 @@ def help(*args)
         super(*args)
       end
 
-      def app_option(*args, &block)
```
ekohl: I think this is still called in hook_context's `app_config`.

lib/kafo/kafo_configure.rb
```diff
@@ -157,46 +207,14 @@ def execute
...
-      builder_class = kafo.full_help? ? HelpBuilders::Advanced : HelpBuilders::Basic
-      args.push builder_class.new(kafo.params)
-      super(*args)
-    end
-
-    def self.app_option(*args, &block)
-      self.app_options ||= []
-      self.app_options.push self.option(*args, &block)
-      self.app_options.last
-    end
```
ekohl, line 198: This option (or rather, the moved one) is still used in HookContext, which refers to `self.kafo.class.app_option`

ekohl (commented):
Overall this feels like a much cleaner implementation.

lib/kafo/kafo_configure.rb
ehelms: Yep, that's intended not to change, as that corresponds to the defined method here: https://github.com/theforeman/kafo/pull/299/files#diff-0598f0904f7f79914887ee5bf9fee84f5f21e599bdba5e885a6d46b9da703505R9 which is included in KafoConfigure making the option available.

lib/kafo/kafo_configure.rb
```diff
@@ -372,7 +395,7 @@ def is_value?(str)
     end
 
     def parse_app_arguments
-      self.class.app_options.each do |option|
+      self.class.declared_options.each do |option|
```
ekohl, line 398: I think this may introduce a regression where it stores all the params in `config.app` as well. Perhaps you can filter with `is_a?(AppOption)` on `declared_options`? Or introduce a compat helper since it's repeated above in `clamp_app_arguments`:
```ruby
def app_options
  declared_options.select { |option| option.is_a?(Kafo::AppOption::Definition) }
end
```

ehelms:
That is what https://github.com/theforeman/kafo/pull/292 should fix and is
the right place to do so given that's the code that controls and defines
what is saved.

ekohl:
I disagree with that. Not all params should end up on the app options hash. At runtime you have different methods of retrieving that data and it can be confusing, leading to bugs. For example, `app_option?` from the hook context will behave different:
https://github.com/theforeman/kafo/blob/9fb119aff2f9827ecd088438aacc613f3c56c8c1/lib/kafo/hook_context.rb#L41-L47

I think it's important to only have the really relevant data present. A developer may not read the documentation and start to modify params in the wrong place using the wrong method. Please keep it compatible.

lib/kafo/kafo_configure.rb
ehelms: First, *all* options are AppOption objects with this change. We can introduce a ModuleOption to be pedantic if we want but it's functionality would not vary. Secondly, there is no regression here due to the order in which things are processed and happen. The app arguments are parsed and handled before the params (module parameters are) dealt with so that they never end up in the config.app object.

https://github.com/theforeman/kafo/pull/299/files#diff-ef4f6d9e6e9975162ff8026e770e8787730a3547080b75534b8bdb411ff6efc2R145

Then at the end of initialization the module options are created:

https://github.com/theforeman/kafo/pull/299/files#diff-ef4f6d9e6e9975162ff8026e770e8787730a3547080b75534b8bdb411ff6efc2R151

```
[42, 51] in /root/kafo/lib/kafo/hook_context.rb
   42:     # determined during boot; this helper can be used in later hooks to determine whether the option exists.
   43:     #
   44:     # @param [Symbol, String] option
   45:     def app_option?(option)
   46:       require 'byebug';byebug
=> 47:       self.kafo.config.app.key?(option.to_sym)
   48:     end
   49: 
   50:     # @param [Symbol, String] option
   51:     #
(byebug) self.kafo.config.app
{:name=>"Foreman", :description=>"Default installation of Foreman", :enabled=>true, :log_dir=>"./_build/", :store_dir=>"", :log_name=>"foreman.log", :log_level=>:debug, :no_prefix=>false, :mapping=>{}, :answer_file=>"./config/foreman-answers.yaml", :installer_dir=>".", :module_dirs=>"./_build/modules", :colors=>true, :color_of_background=>:dark, :hook_dirs=>[], :custom=>{}, :facts=>{}, :low_priority_modules=>["foreman_proxy_plugin", "foreman_compute", "foreman_plugin"], :verbose_log_level=>"notice", :skip_puppet_version_check=>false, :hiera_config=>"./config/foreman-hiera.yaml", :parser_cache_path=>"./_build/parser_cache/foreman.yaml", :order=>["foreman", "foreman_proxy", "puppet"], :reset_data=>false, :detailed_exitcodes=>false, :clear_pulp_content=>false, :dont_save_answers=>false, :ignore_undocumented=>false, :interactive=>nil, :noop=>false, :profile=>false, :skip_checks_i_know_better=>false, :verbose=>true, :scenario=>nil, :disable_scenario=>nil, :enable_scenario=>nil, :list_scenarios=>nil, :force=>nil, :compare_scenarios=>nil, :migrations_only=>nil, :parser_cache=>nil, :help=>nil}
```

ekohl: How did you call it? Did you pass in `--mymodule-param value` while doing so? I'll try to set up a test env to actually try this patch (rather than just reading the patch).

ehelms: Correct. I did. If you look at the ordering in `initialize` method, the module options are not defined until *after* the app options are both defined and set.

App options defined: https://github.com/theforeman/kafo/pull/299/files#diff-ef4f6d9e6e9975162ff8026e770e8787730a3547080b75534b8bdb411ff6efc2R141
App options set: https://github.com/theforeman/kafo/pull/299/files#diff-ef4f6d9e6e9975162ff8026e770e8787730a3547080b75534b8bdb411ff6efc2R145
Module options defined and values set: 
  * https://github.com/theforeman/kafo/pull/299/files#diff-ef4f6d9e6e9975162ff8026e770e8787730a3547080b75534b8bdb411ff6efc2R151
  * https://github.com/theforeman/kafo/pull/299/files#diff-ef4f6d9e6e9975162ff8026e770e8787730a3547080b75534b8bdb411ff6efc2R357-R363

ekohl: To make things easier to test, I wrote https://github.com/ekohl/kafo-example-installer. Using this I confirmed this.

ekohl: > We can introduce a ModuleOption to be pedantic if we want but it's functionality would not vary.

I wonder if it makes sense in a follow up patch.

lib/kafo/kafo_configure.rb
```diff
@@ -28,12 +28,15 @@ module ENV
...
 
 module Kafo
   class KafoConfigure < Clamp::Command
     include StringHelper
 
     class << self
+      include AppOption::Declaration
+
       attr_accessor :config, :root_dir, :config_file, :gem_root,
                     :module_dirs, :kafo_modules_dir, :verbose, :app_options, :logger,
```
ekohl, line 41: You should drop `app_options` here as well.
```suggestion
                    :module_dirs, :kafo_modules_dir, :verbose, :logger,
```

ekohl (commented):
Other than the one minor inline comment, I think this looks complete. Tested and confirmed it works.

lib/kafo/kafo_configure.rb
```diff
@@ -366,13 +367,15 @@ def set_options
     # so we accept either allowed args or those that does not start with '-' and are right after
     # accepted argument
     def clamp_app_arguments
-      @allowed_clamp_app_arguments = self.class.app_options.map do |option|
+      @allowed_clamp_app_arguments = self.class.declared_options.map do |option|
         option.switches.map { |s| is_yes_no_flag?(s) ? build_yes_no_variants(s) : s }
```
ekohl, line 371: Not for this patch but reading this and the design I started to wonder if we should call the private method [recognised_switches](https://github.com/mdub/clamp/blob/b95498a2c44c48d1703548dc20625ec528b97166/lib/clamp/option/definition.rb#L74-L82) instead. Now that we have defined our own class, we can expose it and drop our copies.

ekohl (approved):

lib/kafo/kafo_configure.rb
ehelms: Going to track that suggestion here so I do not forget about it -- https://projects.theforeman.org/issues/31175

PR #301, opened 2020-10-27
Add ability to declare modules that can be disabled

ehelms:
Modules in an answer file are assumed to be able to be enabled
and subsequently disabled. However, turning off a puppet module
entirely can produce negative affects for users depending on the
scenario.

This adds the ability to declare specifically the set of modules
that can be disabled. If any are declared (or an empty list) then
the available app options are adjusted. Any disabled module can be
will get a --enable option, any currently enabled option will not
have any option for enable or disable and any module declared in
the disablable modules list will get a --[no-]enable option for
turning it on and off.

### DIFF
--- a/lib/kafo/configuration.rb
+++ b/lib/kafo/configuration.rb
@@ -30,7 +30,8 @@ module Kafo
         :facts                => {},
         :low_priority_modules => [],
         :verbose_log_level    => 'notice',
-        :skip_puppet_version_check => false
+        :skip_puppet_version_check => false,
+        :disablable_modules   => nil,
     }
 
     def self.get_scenario_id(filename)
--- a/lib/kafo/kafo_configure.rb
+++ b/lib/kafo/kafo_configure.rb
@@ -347,12 +347,7 @@ module Kafo
         request_help
       end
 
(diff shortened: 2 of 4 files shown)


ekohl:
FYI: In https://community.theforeman.org/t/defaulting-puppet-to-off-in-the-katello-scenario/14553 I wrote a suggestion to restructure the answers file to allow storing it there. I'd like to hear your thoughts on why you think it's better to store it as part of the scenario as opposed to the answers file.

ehelms:
> FYI: In https://community.theforeman.org/t/defaulting-puppet-to-off-in-the-katello-scenario/14553 I wrote a suggestion to restructure the answers file to allow storing it there. I'd like to hear your thoughts on why you think it's better to store it as part of the scenario as opposed to the answers file.

From your design, you'd be thinking to change the layout to:

```
module:
  enabled: false
  disablable: true
  parameters:
```

I think that design would work for me. That will require a much larger change overall to the structure and I think the work to move the answers file to be treated like a database. I attempted to write this change in a way that it adds functionality but does not require a bigger change. I think that's it's only real advantage at this point.

PR #305, opened 2020-11-05
Fixes #31269: Roll logs by date

ehelms:
After more than 5-10 runs of an installer, the log files lose all semblance of when they were run just looking at the log files. Someone debugging has to look inside the log files and piece together the timeline themselves. Switching to rolling by date and timestamp provides a clearer view of the order of log files, when each run of the installer was executed to pair with other system logs and how many times the installer was ran in some window of time.

### DIFF
--- a/lib/kafo/logging.rb
+++ b/lib/kafo/logging.rb
@@ -46,7 +46,8 @@ module Kafo
             level: log_level,
             filename: filename,
             layout: layout(color: false),
-            truncate: true
+            truncate: true,
+            roll_by: 'date'
           )
 
           FileUtils.chown(


wbclark (approved):
I love the idea of what this accomplishes. It's indeed frustrating when debugging to see `/var/log/foreman-installer/katello.{1..20}.log` and have to investigate further to determine which is relevant.

I understand from the Logger documentation that `roll_by: 'date'` includes the date in the filename. That's a big win from my perspective.

lib/kafo/logging.rb
```diff
@@ -46,7 +46,8 @@ def setup_file_logging(log_level, log_dir, log_owner, log_group)
             level: log_level,
             filename: filename,
             layout: layout(color: false),
-            truncate: true
```
ekohl, line 49: Every time I see diffs like this I'm convinced we should use more trailing commas.

ekohl (commented):
:+1: for dates. Another benefit from the readability point of view is that you don't need to rotate all existing files which is much more efficient from a storage perspective.

PR #307, opened 2020-11-11
Fixes #31304: set exit code before post hooks

ehelms:
If the exit code is not set until the exit handler runs, post hooks
can incorrectly surmise the current exit code. The puppet run sets
the exit code after execution. Users of Kafo such as the foreman-installer
rely on the exit_code in some of the post hooks to determine behavior.

### DIFF
--- a/test/fixtures/hooks/post/10-exit-code.rb
+++ b/test/fixtures/hooks/post/10-exit-code.rb
@@ -1,0 +1,1 @@
+puts @kafo.exit_code
--- a/lib/kafo/kafo_configure.rb
+++ b/lib/kafo/kafo_configure.rb
@@ -465,7 +465,7 @@ module Kafo
         'show_diff' => true,
       )
 
-      exit_code   = 0
+      self.class.exit_handler.exit_code = 0
       exit_status = nil
       options     = [
           '--verbose',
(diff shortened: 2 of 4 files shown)


ekohl (changes requested):
Please explain in a commit message why this makes sense.

ehelms:
[test kafo]

ehelms:
Tests continue to fail -- proposing https://github.com/theforeman/kafo/pull/308 to address and unblock PRs

ehelms:
Failures caused by https://github.com/theforeman/foreman-infra/pull/1515

ehelms:
[test kafo]

ehelms:
[test kafo]

ehelms:
[test kafo]

ehelms:
[test kafo]

ehelms:
Jenkins is cleaned up and tests passing now, we will need this for Foreman 2.3.

wbclark (commented):
Post install and upgrade hooks rely on checking `kafo.exit_code` and with this PR we now set it globally after puppet apply.

This feels like something that should have a test so that it isn't easy to break in the future?

ehelms:
[test kafo]

ehelms:
@wbclark Test added, let me know what you think.

test/fixtures/hooks/post/10-exit-code.rb
```diff
@@ -0,0 +1 @@
+puts @kafo.exit_code
```
wbclark, line 1: cute :)

wbclark (approved):

PR #311, opened 2020-11-18
Fixes #31332: Do not display log output when using --help

ehelms:
Logs are buffered before logging configuration is set up, such that
once they are they are added to the log file and to the terminal
output when using verbose mode. That could be triggered when
running --help causing the user to see some log output which
is confusing when using --help. This skips log setup if --help
is present and also prevents new log files from being created
unnecessarily.

### DIFF
--- a/lib/kafo/kafo_configure.rb
+++ b/lib/kafo/kafo_configure.rb
@@ -143,7 +143,7 @@ module Kafo
       # so we limit parsing only to app config options (because of --help and later defined params)
       parse clamp_app_arguments
       parse_app_arguments # set values from ARGS to config.app
-      Logging.setup(verbose: config.app[:verbose])
+      Logging.setup(verbose: config.app[:verbose]) unless ARGV.any? { |option| ['--help', '--full-help'].include? option }
       self.class.set_color_scheme
 
       self.class.hooking.execute(:init)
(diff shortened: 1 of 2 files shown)


wbclark (commented):
This change only modifies `test/acceptance/kafo_configure_test.rb` at this point, adding additional checks that `--{,full-}help` output neither includes `NOTICE` nor `Executing hooks`.

My expectation is that CI should have failed because you didn't yet change actual logic in KafoConfigure, but the tests actually succeeded. What am I missing?

ehelms:
Doh, somehow the code change did not get committed. And what you point is a good catch, since verbose is not default these pass. I will update the tests to reflect that you have to be using verbose mode with help.

ehelms:
OK, now it has the code change and proper tests. Thanks for pointing that out @wbclark

lib/kafo/kafo_configure.rb
```diff
@@ -143,7 +143,7 @@ def initialize(*args)
       # so we limit parsing only to app config options (because of --help and later defined params)
       parse clamp_app_arguments
       parse_app_arguments # set values from ARGS to config.app
-      Logging.setup(verbose: config.app[:verbose])
+      Logging.setup(verbose: config.app[:verbose]) unless ARGV.include?('--help') || ARGV.include?('--full-help')
```
wbclark, line 146:
```suggestion
      Logging.setup(verbose: config.app[:verbose]) unless ARGV.any? { |option| ['--help', '--full-help'].include? option }
```

I think we may want to have some Hooks suppress this output as well. This is logically equivalent, but easier to expand on.

ekohl:
https://github.com/theforeman/foreman-installer/pull/610 also suffers from this. I guess this is why in the past logging was not set up until a later stage.

wbclark (approved):

PR #312, opened 2020-11-18
Fixes #30365 - Resolve symlinks to determine AIO Puppet

ekohl:
Commit 695ad18f7f3dccb07d93604701f68345c665b181 attempted to deal with symlinks in /usr/bin or /usr/local/bin, but broke things because paths in /opt/puppetlabs/bin are also symlinks. That's why it was reverted in eacdf43c780ef5ac7fab90f0ce3d49b04539bba3.

This uses File.realpath to resolve symlinks so /usr/local/bin/puppet is fine as long as it is pointing to a file in /opt/puppetlabs.

I haven't had a chance to test this PR.

### DIFF
--- a/lib/kafo/puppet_command.rb
+++ b/lib/kafo/puppet_command.rb
@@ -43,8 +43,15 @@ module Kafo
       File.join([bin_path, bin_name].compact)
     end
 
+    def self.is_aio_puppet?
+      puppet_command = search_puppet_path('puppet')
+      File.realpath(puppet_command).start_with?('/opt/puppetlabs')
+    rescue Errno::ENOENT
+      false
+    end
+
     def self.format_command(command)
-      if search_puppet_path('puppet').start_with?('/opt/puppetlabs')
+      if is_aio_puppet?
         [clean_env_vars, command, :unsetenv_others => true]
       else
         [::ENV, command, :unsetenv_others => false]
--- a/test/kafo/puppet_command_test.rb
+++ b/test/kafo/puppet_command_test.rb
@@ -81,5 +81,45 @@ module Kafo
         specify { File.stub(:executable?, false) { _(pc).must_equal 'puppet' } }
       end
     end
+
+    describe '.is_aio_puppet?' do
+      subject do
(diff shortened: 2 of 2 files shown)


lib/kafo/puppet_command.rb
```diff
@@ -40,7 +40,7 @@ def self.search_puppet_path(bin_name)
       bin_path = (::ENV['PATH'].split(File::PATH_SEPARATOR) + ['/opt/puppetlabs/bin']).find do |path|
         File.executable?(File.join(path, 'puppet')) && File.executable?(File.join(path, bin_name))
       end
-      File.join([bin_path, bin_name].compact)
+      bin_path ? File.realpath(File.join(bin_path, bin_name)) : bin_name
```
ehelms, line 43: From my testing, this will not unfortunately work. The way the Puppet AIO wrapper code works, it is not intended to be called directly. So the symlinks resolve the wrapper, but the wrapper expects that and if it is called on itself it enters an infinite loop.

For example, puppet bin:
```
/opt/puppetlabs/bin/puppet -> /opt/puppetlabs/puppet/bin/wrapper.sh
```

Which this new code will resolve to return `/opt/puppetlabs/puppet/bin/wrapper.sh` as the bin file. However, that wrapper code does the following:

```
COMMAND=`basename "${0}"`

exec /opt/puppetlabs/puppet/bin/${COMMAND} "$@"
```

Which, if you call the `wrapper.sh` directly, `COMMAND` resolves to `wrapper.sh` and it will keep calling itself endlessly.

ekohl: Ok thanks for that. Sounds like we should perform `File.realpath` and see if **that** is in `/opt/puppetlabs`, but still use the original path we found.

ekohl:
How about this version? If you have a test env at hand I'd also appreciate if you could give it a spin.

ehelms (approved):
Worked like a charm for me.

This I think should get a similar treatment -- https://github.com/theforeman/kafo_parsers/blob/master/lib/kafo_parsers/puppet_strings_module_parser.rb#L137

PR #314, opened 2020-12-07
Fixes #31466: Switch HighLine back to using :info color

ehelms:
HighLine does not have a :notice color which was changed as part
of the switchover from info to notice as the default log level.

### DIFF
--- a/test/acceptance/kafo_configure_test.rb
+++ b/test/acceptance/kafo_configure_test.rb
@@ -52,6 +52,15 @@ module Kafo
       end
     end
 
+    describe '--list-scenarios' do
+      it 'lists scenarios' do
+        code, out, err = run_command '../bin/kafo-configure --list-scenarios --no-colors'
+        _(code).must_equal 0, err
+        _(out).must_include "Available scenarios"
+        _(out).must_include "default (use: --scenario default)"
+      end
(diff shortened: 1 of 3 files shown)


ekohl (commented):
Should we add some tests to forklift that calls foreman-installer --list-scenarios to catch these regressions in the future?

ehelms:
I pushed a test that handles checking it right here in the Kafo code base.

test/acceptance/kafo_configure_test.rb
```diff
@@ -52,6 +52,15 @@ module Kafo
       end
     end
 
+    describe '--list-scenarios' do
+      it 'lists scenarios' do
+        code, out, err = run_command '../bin/kafo-configure --list-scenarios'
+        _(code).must_equal 0, err
+        _(out).must_include "Available scenarios"
+        _(out).must_include "default \e[0m(use: --scenario default)\n"
```
evgeni, line 60: seems the tests don't use colors and fail on the difference.

ehelms: Darn dynamic calculation based on environment. I set it to `--no-colors` to be explicit.

evgeni (approved):

PR #322, opened 2021-03-09
Refs #31779 - Print information about log based output mode

wbclark:
We should inform the user about log based terminal output mode and the available verbosity levels, and this should happen sooner rather than waiting for the system configuration step.

### DIFF
--- a/lib/kafo/kafo_configure.rb
+++ b/lib/kafo/kafo_configure.rb
@@ -166,6 +166,10 @@ module Kafo
     def run(*args)
       started_at = Time.now
       logger.debug("Running installer with args #{args.inspect}")
+      if config.app[:verbose]
+        logger.notice("Running installer with log based terminal output at level #{config.app[:verbose_log_level].upcase}.")
+        logger.notice("Use -l to set the terminal output log level to ERROR, WARN, NOTICE, INFO, or DEBUG.")
+      end
       super
     ensure
       logger.debug("Installer finished in #{Time.now - started_at} seconds")
@@ -477,13 +481,7 @@ module Kafo
         log_parser = PuppetLogParser.new
         logger = Logger.new('configure')
 
-        start_message = <<-HEREDOC
-Starting system configuration.
-  The total number of configuration tasks may increase during the run.
-  Observe logs or use `--verbose-log-level info` to see individual configuration tasks.
-HEREDOC
-
-        logger.notice(start_message.chomp)
+        logger.notice("Starting system configuration.")
 
         PTY.spawn(*PuppetCommand.format_command(command)) do |stdin, stdout, pid|
           begin


lib/kafo/kafo_configure.rb
```diff
@@ -166,6 +166,12 @@ def logger
     def run(*args)
       started_at = Time.now
       logger.debug("Running installer with args #{args.inspect}")
+      if config.app[:verbose]
+        log_output_message = <<~MESSAGE.strip
+          Running installer with log based terminal output at level #{config.app[:verbose_log_level].upcase}. Use -l to change output verbosity to one of ERROR, WARN, NOTICE, INFO, or DEBUG.
```
ehelms, line 171: Does the option handle uppercase log levels? I feel like it should but I cannot say I have ever tried it.

wbclark: Yes, it works fine in my testing to run `foreman-installer -l INFO` for example

ehelms (approved):

lib/kafo/kafo_configure.rb
ekohl: Looks like it's normalized here:
https://github.com/TwP/logging/blob/ae9872d093833b2a5a34cbe1faa4e895a81f6845/lib/logging.rb#L513

ekohl, line 171: Small nit: first you speak of `level X` and then about `output verbosity`. Would it make sense to always speak of level? I also think log levels are a well known concept any admin should understand while `output verbosity` may not. Something like
```suggestion
          Running installer with log based terminal output at level #{config.app[:verbose_log_level].upcase}. Use -l to set the level to ERROR, WARN, NOTICE, INFO, or DEBUG.
```

I'd also consider adding a newline after the dot to stay somewhat close to the max line length.

lib/kafo/kafo_configure.rb
```diff
@@ -166,6 +166,12 @@ def logger
...
+        MESSAGE
+        logger.notice(log_output_message)
```
ekohl, line 173: Sending this to the logger feels weird. It also means it ends up in the log files where it doesn't provide any value. Perhaps it should use regular `puts`?

ekohl (commented):
Small nits inline that I'd like you to think about but I think the general place of where it's printed makes sense. Also the general intent of the message looks like an improvement.

I also hoped we could output the valid choices in `--help` but it looks like clamp doesn't support that out of the box. See https://docs.python.org/3/library/argparse.html#choices for what I was thinking about.

wbclark:
> I also hoped we could output the valid choices in `--help` but it looks like clamp doesn't support that out of the box.

I opened https://github.com/theforeman/kafo/pull/323 to do just that. Your feedback would be valuable there as well

lib/kafo/kafo_configure.rb
wbclark: Done. Thanks for the suggestion!

lib/kafo/kafo_configure.rb
```diff
@@ -166,6 +166,13 @@ def logger
     def run(*args)
       started_at = Time.now
       logger.debug("Running installer with args #{args.inspect}")
+      if config.app[:verbose]
+        log_output_message = <<~MESSAGE.strip
+          Running installer with log based terminal output at level #{config.app[:verbose_log_level].upcase}.
+          Use -l to set the terminal output log level to ERROR, WARN, NOTICE, INFO, or DEBUG.
+        MESSAGE
+        logger.notice(log_output_message)
```
ekohl, line 174: I don't know how well our tooling deals with multi-line output. Perhaps it should be this:
```suggestion
        puts "Running installer with log based terminal output at level #{config.app[:verbose_log_level].upcase}."
        puts "Use -l to set the terminal output log level to ERROR, WARN, NOTICE, INFO, or DEBUG."
```

lib/kafo/kafo_configure.rb
wbclark: I don't think it adds any confusion if this message additionally shows up in the logs. The more common case is that the user will see it in their terminal output (under most common circumstances it will be the first thing they see). That gets me thinking,

1. Is it desired to see the time, stage, and log level for this message? I think that's mostly an aesthetic choice.

2. Should this message be displayed even if using `-l ERROR` for example? That would be an advantage of using `puts`, for the price of making the visual formatting of the output less consistent. But how common is that, and isn't the point of `-l ERROR` to not show output unless it's a problem?

lib/kafo/kafo_configure.rb
wbclark: I want to see for myself how the actual output looks with these approaches we are discussing. I'll test it and report back shortly

lib/kafo/kafo_configure.rb
ekohl: That's an interesting point. If the user passed in `-l` (or to be more precise, the value for `-l` is different from the default), do we need to inform the user about the `-l` option anyway? This goes both ways, so with `-l DEBUG` we can also leave it out.

lib/kafo/kafo_configure.rb
```diff
@@ -166,6 +166,13 @@ def logger
     def run(*args)
       started_at = Time.now
       logger.debug("Running installer with args #{args.inspect}")
+      if config.app[:verbose]
+        log_output_message = <<~MESSAGE.strip
+          Running installer with log based terminal output at level #{config.app[:verbose_log_level].upcase}.
+          Use -l to set the terminal output log level to ERROR, WARN, NOTICE, INFO, or DEBUG.
```
wbclark, line 172:
```suggestion
          Use -l to set the terminal output log level to ERROR, WARN, NOTICE, INFO, or DEBUG.
          Or use --progress-bar to display a simple progress bar instead of log based terminal output.
```

@ehelms @ekohl thoughts on including this as well?

ehelms: Given the progress bar is broken, and we were tyring to make a more permanent move to the log based output. I would opt for keeping that part out.

lib/kafo/kafo_configure.rb
ehelms: Personally, I like sticking with log based output for the consistency of it visually and code wise.

Out of curiosity, is anyone aware of other tools that do the above that we are trying to do here? I'm not saying we have to adhere to trends, but I tend to think about comparing this to what other experiences users might have and their experiences therein.

wbclark: Thanks for the feedback. Based on this discussion, I think the most visually consistent approach is to keep using `logger.notice` but split it into two lines. The latest change I pushed takes that approach.

lib/kafo/kafo_configure.rb
wbclark: That sounds good. Based on conversation today on this topic, we may want to discuss dropping the progress bar entirely.

ehelms (approved):

lib/kafo/kafo_configure.rb
```diff
@@ -477,13 +481,7 @@ def run_installation
         log_parser = PuppetLogParser.new
         logger = Logger.new('configure')
 
-        start_message = <<-HEREDOC
-Starting system configuration.
-  The total number of configuration tasks may increase during the run.
```
ekohl, line 482: I just noticed that we no longer print this but it was something that surprised me the first time I saw it. Does it still make sense to warn the user about this?

wbclark: Usually, the increase in the total number of configuration tasks is small relative to the original reported total (less than 1% delta in my testing). I reason that because the delta is small, and because the concept is intuitive enough that one configuration might trigger another previously unplanned item, we could save the line of output.

If we want to make this clearer, I think we can do so without causing additional lines of output, with language like

`2021-03-12 15:11:10 [NOTICE] [configure] 250 configuration steps out of 2188 currently planned steps complete.`

Still, it reads a little bit clunky to me, and probably unnecessary?

PR #323, opened 2021-03-11
Fixes #32074 - Document output log levels in --help output

wbclark:


ehelms:
Seems reasonable to me, this will make the current already verbose help more verbose unfortunately. The tests would need to be updated. I'd like to see what @ekohl thinks.

wbclark:
Tests are fixed.

It would be helpful for someone to review the language describing each log level and look for improvements.

ehelms:
This is definitely more output about log levels than I have encountered in a CLI program before. I'm torn on how to feel about it. It's useful information just verbose and expands are already large help text even further. I feel like you usually see this in a config file, a man page or the online documentation. To that end, I am torn. If this is valuable to users at this point in the help docs, I am OK with it. If this is overly verbose and unexpected by users, I'm inclined to include this information in another way. 

@ekohl @evgeni I think we need some additional voices here.

lib/kafo/kafo_configure.rb
```diff
@@ -338,7 +338,15 @@ def set_app_options
                  :default => false, :advanced => true
       app_option ['-v', '--[no-]verbose'], :flag, 'Display log on STDOUT instead of progressbar',
                  :default => config.app[:verbose]
-      app_option ['-l', '--verbose-log-level'], 'LEVEL', 'Log level for verbose mode output',
+      log_levels = <<~HEREDOC.chomp
+        Log level for log based terminal output. The available levels are
+        ERROR  - Only show errors which prevented the installer from completing successfully.
+        WARN   - Deprecation warnings and other information users may want to be aware of.
+        NOTICE - High level information about installer execution and progress.
```
ekohl, line 345: I think this is incorrect. https://en.wikipedia.org/wiki/Syslog#Severity_level has a nice table and describes notice as:
> Normal but significant conditions. Conditions that are not error conditions, but that may require special handling.

ekohl (commented):
It feels quite long and in general I'd avoid multi line as much as possible since it can get very odd with indenting etc. I am wondering if we can use `--help` vs `--full-help` here as well.

wbclark:
> It feels quite long and in general I'd avoid multi line as much as possible since it can get very odd with indenting etc. I am wondering if we can use `--help` vs `--full-help` here as well.

I would say most users are unfortunately not aware of `--full-help` and even if they are you still need to use a tool like `less` or `grep` to usefully review its output in your terminal since the volume of output is so high. It feels like a place where help text goes to die, unless someone is really committed to digging for a specific thing. I think the likelihood of users learning something valuable about log levels from `--full-help` output is unfortunately quite low.

lib/kafo/kafo_configure.rb
wbclark: The same resource also states, "the meaning of severity levels other than Emergency and Debug are relative to the application" which I would argue is applicable in this case. The user's first impression of [NOTICE] level will now be the terminal output they see when running the installer, which typically doesn't suggest or require much special handling. "Normal but significant conditions" is perhaps more in line with how we are using the level.... the description I proposed here is like a version of "normal but significant conditions" but worded specifically for the installer.

evgeni:
I am torn. This is a lot of output for `--help` and you'd *think* people would know the meaning of the (rather standardized) log level names. And yeah, too few people know about `--full-help`, but maybe that's a different thing we should fix?

Anyways, can we have different output for `--help` vs `--full-help`? Like this:

```console
# foreman-installer --help
…
-l, --verbose-log-level -- Log level for log based terminal output.
        The available levels are ERROR, WARN, NOTICE, INFO, DEBUG. See the full-help for details.
# foreman-installer --full-help
…
-l, --verbose-log-level -- Log level for log based terminal output.
        The available levels are
        ERROR  - Only show errors which prevented the installer from completing successfully.
        WARN   - Deprecation warnings and other information users may want to be aware of.
        NOTICE - High level information about installer execution and progress.
        INFO   - More detailed information about execution and progress. Also shows when the installer makes a change to system configuration.
        DEBUG  - Show all information about execution, including configuration items where no change was needed.
```

(plus proper formatting)

mccun934:
I like @evgeni 's idea of the call-out to using -full-help for more details. If this isn't too much extra work I'd go with his suggested formatting

wbclark:
Updated with my (right now, very ugly) attempt at making this work.

I was pleasantly surprised to find that clamp doesn't complain if I declare the same option multiple times in advanced/non-advanced versions.

The issue I am facing is hiding the ordinary --help output when --full-help is run.

wbclark:
Updated with a cleaner solution that actually works.

lib/kafo/kafo_configure.rb
```diff
@@ -316,6 +316,25 @@ def app_option(*args, &block)
       self.class.app_option(*args, &block)
     end
 
+    def terminal_log_levels_message
+      if ARGV.include?('--full-help')
```
evgeni, line 320: I think you can use `kafo.full_help?` here instead, which is cleaner.

evgeni: aaaaaaaaah, yeah, THAT is possible. you're still initing clamp here, so it didn't parse stuff. you're right.

wbclark:
@ehelms `--help` vs. `--full-help` output as requested:

```
# foreman-installer --help
Usage:
    foreman-installer [OPTIONS]
 
Options:
 
= Generic:
    --[no-]lock-package-versions  Let installer lock versions of the installed packages to prevent
                                  unexpected breakages during system updates. The choice is remembered
                                  and used in next installer runs. (default: nil)
    --tuning INSTALLATION_SIZE    Tune for an installation size. Choices: default, medium, large, extra-large, extra-extra-large (default: "default")
    --certs-update-server         This option will enforce an update of the HTTPS certificates (default: false)
    --certs-update-server-ca      This option will enforce an update of the CA used for HTTPS certificates. (default: false)
    --certs-update-all            This option will enforce an update of all the certificates for given host (default: false)
    --certs-reset                 This option will reset any custom certificates and use the self-signed CA instead. Note that any clients will need to be updated with the latest katello-ca-consumer RPM, and any external proxies will need to have the certs updated by generating a new certs tarball. (default: false)
    --certs-skip-check            This option will cause skipping the certificates sanity check. Use with caution (default: false)
    -i, --interactive             Run in interactive mode
    -n, --noop                    Run puppet in noop mode? (default: false)
    -s, --skip-checks-i-know-better Skip all system checks (default: false)
    -v, --[no-]verbose            Display log on STDOUT instead of progressbar (default: true)
    -l, --verbose-log-level LEVEL Log level for log based terminal output.
                                  The available levels are ERROR, WARN, NOTICE, INFO, DEBUG. See --full-help for definitions. (default: "notice")
    --list-scenarios              List available installation scenarios
    -h, --help                    print help
    --full-help                   print complete help
    --[no-]enable-certs           Enable 'certs' puppet module (default: true)
    --[no-]enable-foreman         Enable 'foreman' puppet module (default: true)
....
```
vs.
```
# foreman-installer --help
Usage:
            foreman-installer [OPTIONS]
         
        Options:
         
        = Generic:
            --[no-]lock-package-versions  Let installer lock versions of the installed packages to prevent
                                          unexpected breakages during system updates. The choice is remembered
                                          and used in next installer runs. (default: nil)
            --reset-data                  This option will drop all databases for Foreman and subsequent backend systems. You will lose all data!
                                          Unfortunately, we can't detect a failure, so you should verify success manually.
                                          Dropping can fail when the DB is in use. (default: false)
            --detailed-exitcodes          Provide transaction information via exit codes, see puppet-agent(8)
                                          for full details. (default: false)
            --clear-pulp-content          This option will clear all Pulp 2 data. (default: false)
            --tuning INSTALLATION_SIZE    Tune for an installation size. Choices: default, medium, large, extra-large, extra-extra-large (default: "default")
            --certs-update-server         This option will enforce an update of the HTTPS certificates (default: false)
            --certs-update-server-ca      This option will enforce an update of the CA used for HTTPS certificates. (default: false)
            --certs-update-all            This option will enforce an update of all the certificates for given host (default: false)
            --certs-reset                 This option will reset any custom certificates and use the self-signed CA instead. Note that any clients will need to be updated with the latest katello-ca-consumer RPM, and any external proxies will need to have the certs updated by generating a new certs tarball. (default: false)
            --certs-skip-check            This option will cause skipping the certificates sanity check. Use with caution (default: false)
            --disable-system-checks       This option will skip the system checks for memory. (default: false)
            --[no-]colors                 Use color output on STDOUT (default: true)
            --color-of-background COLOR   Your terminal background is :bright or :dark (default: :dark)
            --dont-save-answers           Skip saving answers to '/etc/foreman-installer/scenarios.d/katello-answers.yaml'? (default: false)
            --ignore-undocumented         Ignore inconsistent parameter documentation (default: false)
            -i, --interactive             Run in interactive mode
            --log-level LEVEL             Log level for log file output (default: "DEBUG")
            -n, --noop                    Run puppet in noop mode? (default: false)
            -p, --profile                 Run puppet in profile mode? (default: false)
            -s, --skip-checks-i-know-better Skip all system checks (default: false)
            --skip-puppet-version-check   Skip check for compatible Puppet versions (default: false)
            -v, --[no-]verbose            Display log on STDOUT instead of progressbar (default: true)
            -l, --verbose-log-level LEVEL Log level for log based terminal output.
                                          The available levels are
                                          ERROR  - Only show errors which prevented the installer from completing successfully.
                                          WARN   - Deprecation warnings and other information users may want to be aware of.
                                          NOTICE - High level information about installer execution and progress.
                                          INFO   - More detailed information about execution and progress. Also shows when the installer makes a change to system configuration.
                                          DEBUG  - Show all information about execution, including configuration items where no change was needed. (default: "notice")
            --disable-scenario SCENARIO   Disable installation scenario
            --enable-scenario SCENARIO    Enable installation scenario
            --list-scenarios              List available installation scenarios
            --force                       Force change of installation scenario
            --compare-scenarios           Show changes between last used scenario and the scenario specified with -S or --scenario argument
            --migrations-only             Apply migrations to a selected scenario and exit
            --[no-]parser-cache           Force use or bypass of Puppet module parser cache
            -h, --help                    print help
            --full-help                   print complete help
            --[no-]enable-certs           Enable 'certs' puppet module (default: true)
            --[no-]enable-foreman         Enable 'foreman' puppet module (default: true)
....
```

ehelms (approved):

PR #324, opened 2021-03-11
Fixes #32088 - Change configuration steps log message frequency and wording

wbclark:
This information should not be relevant to the user except in troubleshooting, so let's log it at level DEBUG which will hide it from the new default log based terminal output in standard operation

### DIFF
--- a/lib/kafo/configuration.rb
+++ b/lib/kafo/configuration.rb
@@ -198,7 +198,7 @@ module Kafo
           }
 EOS
 
-        @logger.notice 'Loading default values from puppet modules...'
+        @logger.info "Loading default values from puppet modules..."
         command = PuppetCommand.new(dump_manifest, [], puppetconf, self).command
         stdout, stderr, status = Open3.capture3(*PuppetCommand.format_command(command))
 
@@ -223,7 +223,7 @@ EOS
           end
         end
 
-        @logger.notice "... finished"
+        @logger.info "... finished loading default values from puppet modules."
 
         load_yaml_from_output(stdout.split($/))
       end
--- a/lib/kafo/kafo_configure.rb
+++ b/lib/kafo/kafo_configure.rb
@@ -144,6 +144,7 @@ module Kafo
       parse clamp_app_arguments
       parse_app_arguments # set values from ARGS to config.app
       Logging.setup(verbose: config.app[:verbose]) unless ARGV.any? { |option| ['--help', '--full-help'].include? option }
+      logger.notice("Loading installer configuration. This will take some time.")
       self.class.set_color_scheme
 
       self.class.hooking.execute(:init)
@@ -440,7 +441,7 @@ module Kafo
     end
 
     def validate_all(logging = true)
-      logger.notice 'Running validation checks'
+      logger.info "Running validation checks."
       results = enabled_params.map do |param|
         result = param.valid?
         errors = param.validation_errors.join(', ')
@@ -493,8 +494,8 @@ HEREDOC
               progress_log(method, message, logger)
 
               if (output = line.match(%r{(.+\]): Starting to evaluate the resource( \((?<count>\d+) of (?<total>\d+)\))?}))
-                if (output[:count].to_i % 100) == 1 && output[:count].to_i != 1
-                  logger.notice("#{output[:count].to_i - 1} out of #{output[:total]} done.")
+                if (output[:count].to_i % 250) == 1 && output[:count].to_i != 1
+                  logger.notice("#{output[:count].to_i - 1} configuration steps out of #{output[:total]} steps complete.")
                 end
               end
 


ehelms:
I know in the past we had discussed keeping this at notice level because of the length of time it takes to execute. The user may not care that specifically puppet module defaults are being loaded, but it can result in "dead air" output.

wbclark:
> I know in the past we had discussed keeping this at notice level because of the length of time it takes to execute. The user may not care that specifically puppet module defaults are being loaded, but it can result in "dead air" output.

The startup time does seem a bit long. With this change, and running the date command first so that you can see the exact time differential:

```
# date && foreman-installer
Thu Mar 11 13:42:59 EST 2021
2021-03-11 13:43:16 [NOTICE] [root] Running installer with log based terminal output at level NOTICE.
Use -l to set the terminal output log verbosity to ERROR, WARN, NOTICE, INFO, or DEBUG.
2021-03-11 13:43:20 [NOTICE] [root] Running validation checks
2021-03-11 13:43:34 [NOTICE] [configure] Starting system configuration.
```

The benefit is that the first output the user sees is `2021-03-11 13:43:16 [NOTICE] [root] Running installer with log based terminal output at level NOTICE.` and the drawback is as you pointed out this took about 17 seconds.

I think it's still worth logging these at level DEBUG, and if we need to add a specialized startup message that says "this may take a minute, better grab a snickers" let's do that without dual purposing the message about loading puppet module defaults.

(I understand we couldn't make the 'Running installer with log based terminal output...' message occur sooner, because it depends on --verbose and --verbose-log-level which haven't been loaded until that point, but correct me if I'm wrong about that @ehelms )

wbclark:
While I'm at it, I think `2021-03-11 13:43:20 [NOTICE] [root] Running validation checks` should be downgraded to INFO at least as well.

wbclark:
> I know in the past we had discussed keeping this at notice level because of the length of time it takes to execute. The user may not care that specifically puppet module defaults are being loaded, but it can result in "dead air" output.

After some hacking, I can make the timing of the output as seen below, using `puts` to get near immediate output where the logger isn't yet available:

```
# date && foreman-installer
Thu Mar 11 18:13:24 EST 2021
Loading installer configuration at 2021-03-11 18:13:25 -0500. This will take some time.
Loading installer configuration at 2021-03-11 18:13:25 -0500. This will take some time.
2021-03-11 18:13:44 [NOTICE] [root] Running installer with log based terminal output at level NOTICE.
```

(Unsure why I'm getting that duplicate line)

OR, if waiting until the logger becomes available, I can do instead:

```
# date && foreman-installer
Thu Mar 11 18:13:24 EST 2021
2021-03-11 18:13:37 [NOTICE] [root] Loading installer configuration. This will take some time.
2021-03-11 18:13:44 [NOTICE] [root] Running installer with log based terminal output at level NOTICE.
```

Which looks a little better visually but is a delay of ~13 seconds before any output is visible in the terminal.

Do you have a preference between these two less than ideal options?

lib/kafo/kafo_configure.rb
```diff
@@ -144,6 +144,7 @@ def initialize(*args)
       parse clamp_app_arguments
       parse_app_arguments # set values from ARGS to config.app
       Logging.setup(verbose: config.app[:verbose]) unless ARGV.any? { |option| ['--help', '--full-help'].include? option }
+      logger.notice("Loading installer configuration. This will take some time...")
```
ehelms, line 147: My feeling is we can do without the ellipsis.  And this just be a statement.

lib/kafo/configuration.rb
```diff
@@ -210,7 +210,7 @@ class { '::kafo_configure::dump_values':
 
           if (version_mismatch = /kafo_configure::puppet_version_failure: (.+?\))/.match(stderr))
             puts version_mismatch[1]
-            puts "Cannot continue due to incompatible version of Puppet. Use --skip-puppet-version-check to disable this check."
+            puts 'Cannot continue due to incompatible version of Puppet. Use --skip-puppet-version-check to disable this check.'
```
ehelms, line 213: Did this change fix an issue?

ehelms: This got me thinking about something I read the other week. From a newer project around standardizing Ruby style. I think there used to be a performance issue, perhaps way back, with single and double quoted strings. Reading on this more and looking at style guides, it seems the consensus these days is there is no performance issue and just use double-quoted strings everywhere:

 * https://github.com/testdouble/standard#standardrb-the-rules
 * https://github.com/github/rubocop-github/blob/master/STYLEGUIDE.md#strings
 * https://anti-pattern.com/always-use-double-quoted-strings-in-ruby

A few resources I was reading. I do realize style takes us down a whole nother Rabbit hole. So I'd encourage not changing unrelated strings style wise to prevent that. I have an item to discuss further.

wbclark: No, I just noticed that we don't need double quotes here since we aren't doing any interpolation. It didn't feel like it deserved a separate PR.

With that said, I did notice there are some rubocops we can fix, so I can drop this for now for clarity and address it with that change later.

lib/kafo/kafo_configure.rb
wbclark: Ok, that works for me.

lib/kafo/configuration.rb
wbclark: Thanks, I guess I had outdated information about single vs. double quotes. Based on this information, I'll update this to use double quotes throughout this PR.

ehelms (approved):
I'd be OK if this was squashed.

wbclark:
Squashed commits

PR #326, opened 2021-03-15
Fixes #32099: Check if an Integer for integer data_type 

ehelms:
This fixes a deprecation warning present in Ruby 2.7:

warning: deprecated Object#=~ is called on Integer; it always returns nil

### DIFF
--- a/lib/kafo/data_types/integer.rb
+++ b/lib/kafo/data_types/integer.rb
@@ -19,7 +19,9 @@ module Kafo
       end
 
       def typecast(value)
-        value =~ /\d+/ ? value.to_i : value
+        Integer(value)
+      rescue TypeError, ArgumentError
+        value
       end
 
       def valid?(input, errors = [])
--- a/test/kafo/data_types/integer_test.rb
+++ b/test/kafo/data_types/integer_test.rb
@@ -17,7 +17,9 @@ module Kafo
       describe "#typecast" do
         it { _(Integer.new.typecast(1)).must_equal 1 }
         it { _(Integer.new.typecast('1')).must_equal 1 }
-        it { _(Integer.new.typecast('1foo')).must_equal 1 }
+        it { _(Integer.new.typecast('1.5')).must_equal '1.5' }
+        it { _(Integer.new.typecast(' 10 ')).must_equal 10 }
+        it { _(Integer.new.typecast('1foo')).must_equal '1foo' }
         it { _(Integer.new.typecast('foo')).must_equal 'foo' }
       end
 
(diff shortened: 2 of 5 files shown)


wbclark (approved):

wbclark:
hmmm,

```
"foo".to_i
=> 0
```

is happening somehow in test/kafo/data_types/hash_test.rb ?

ehelms:
Hrmm, means this method is not used how I thought it was used. I'll take another crack at modifying the original.

ehelms:
Updated. I did change one assertion because it seemed odd to me that "1foo" should be consider an Integer and return "1". If you mix letters and numbers, that ought to be a string. An integer should be clearly an integer.

lib/kafo/data_types/integer.rb
```diff
@@ -19,7 +19,9 @@ def to_s
       end
 
       def typecast(value)
-        value =~ /\d+/ ? value.to_i : value
+        Integer(value)
+      rescue
```
ekohl, line 23: Isn't it better to catch `ArgumentError` explicitly? Catching all exceptions is generally a bad practice.

test/kafo/data_types/integer_test.rb
```diff
@@ -17,7 +17,7 @@ module DataTypes
       describe "#typecast" do
         it { _(Integer.new.typecast(1)).must_equal 1 }
         it { _(Integer.new.typecast('1')).must_equal 1 }
-        it { _(Integer.new.typecast('1foo')).must_equal 1 }
+        it { _(Integer.new.typecast('1foo')).must_equal '1foo' }
```
ekohl, line 20: Another difference is `'1.5'`. Previously that became `1` but is now `'1.5'` since it's not a pure integer. I'd also add a testcase for `' 1 '`. It appears to strip so that is good, but I like covering that in tests.

ehelms:
Additional to this, I think having https://github.com/theforeman/jenkins-jobs/pull/5 in place would help catch these outside of runtime testing.

test/kafo/data_types/integer_test.rb
ehelms: Are you saying we should be converting floats to integers? or arguing that unless its truly an integer (or only an integer inside a string) is when we should cast it?

ekohl: I don't think we should convert it, but we should at least have a test case that explicitly describes the behavior.

ehelms:
This required some additional cleanup to get everything to work properly.

ekohl (approved):
The commit message should also mention Float, other than :+1:

PR #327, opened 2021-03-19
Fixes #32135: Find puppet module parser only when needed

ehelms:
Moves finding of the puppet module parser closer to when it's
actually needed such as no cache found. This saves around 0.5-1
seconds during execution.

### DIFF
--- a/lib/kafo/puppet_module.rb
+++ b/lib/kafo/puppet_module.rb
@@ -25,7 +25,7 @@ module Kafo
       end
     end
 
-    def initialize(identifier, parser = self.class.find_parser, configuration = KafoConfigure.config)
+    def initialize(identifier, parser: nil, configuration: KafoConfigure.config)
       @identifier        = identifier
       @configuration     = configuration
       @name              = get_name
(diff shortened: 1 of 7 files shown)


ehelms:
The testing will require some additional work to re-structure them as they currently relied on this for dependency injection.

lib/kafo/puppet_module.rb
```diff
@@ -25,7 +25,7 @@ def self.find_parser
       end
     end
 
-    def initialize(identifier, parser = self.class.find_parser, configuration = KafoConfigure.config)
```
ekohl, line 28: Perhaps it should keep the parser argument but rather set it to `nil`. Then `if @parser.nil?` it calls `self.class.find_parser`. Since `find_parser` returns `:none` if it isn't found, this should be ok and keep the tests working.

ekohl (commented):
Looks like a good improvement. In production setups it should have a cache and not need a parser.

lib/kafo/puppet_module.rb
ehelms: Updated with this suggestion and moved it out of draft

ekohl (approved):

PR #331, opened 2021-04-13
Fixes #32300 - Do not include hiera configs when loading default values

ehelms:


ehelms:
This highlights and attempts to fix the issue outlined in the discussion here: https://github.com/theforeman/forklift/pull/1331#discussion_r609560529

Because `params` classes are treated special when it comes to looking up default values, the inclusion of hiera look ups of a variable on a `params` class such as in the puppet-katello case described in the above Forklift issue, there can be a break in the ability to do the default look up.

In this case, due to hiera config being present, and performing an include on `katello::params` when the value derived by hiera from:

```
katello::params::enable_katello_agent: "%{lookup('foreman_proxy_content::enable_katello_agent')}"
```

Is supplied to the class, the validation of being a `Boolean` fails because the lookup returns an empty string because in the default value look up case the class `foreman_proxy_content` is not available (only all `params` classes are) and thus sets this to a false value.

This attempts to fix the issue but not including hiera configuration when determining default values from the puppet modules.  We need to ask ourselves: does this change make sense in the whole scheme of things?

Alternative work around for the specific case that brought this up is:

 * don't use katello::params to set this variable
 * supply a default value to the lookup to avoid the broken type validation


I have added additional test fixture updates here to highlight the problem and ensure no regressions in the future should we go this route.

ehelms:
For the supplying a default value, I have tried many different method signatures and none of them appear to work:

```
2021-04-13 01:15:41 [NOTICE] [root] Loading installer configuration. This will take some time.
Could not get default values, check log file at /var/log/foreman-installer/katello.log for more information
2021-04-13 01:15:44 [ERROR ] [root] RUBYLIB=/opt/theforeman/tfm/root/usr/share/gems/gems/kafo-6.3.0/lib/kafo/../..//modules: /opt/puppetlabs/bin/puppet apply /tmp/kafo_installation20210413-30703-j2rl9d/dump_manifest.pp --config=/tmp/kafo_installation20210413-30703-j2rl9d/puppet.conf 
2021-04-13 01:15:44 [ERROR ] [root] Warning: /etc/foreman-installer/custom-hiera.yaml: file does not contain a valid yaml hash
Error: Evaluation Error: Error while evaluating a Function Call, Lookup of key 'katello::params::enable_katello_agent' failed: Syntax error in string: %{alias('foreman_proxy_content::enable_katello_agent', undef, undef, false)} (file: /tmp/kafo_installation20210413-30703-j2rl9d/dump_manifest.pp, line: 119, column: 346) on node pipe-katello-server-nightly-centos7.wareagle.example.com

2021-04-13 01:15:44 [ERROR ] [root] Could not get default values, cannot continue
[root@pipe-katello-server-nightly-centos7 kafo-6.3.0]# vim /usr/share/foreman-installer/config/foreman.hiera/scenario/katello.yaml
[root@pipe-katello-server-nightly-centos7 kafo-6.3.0]# foreman-installer --foreman-proxy-content-enable-katello-agent false
2021-04-13 01:16:23 [NOTICE] [root] Loading installer configuration. This will take some time.
Could not get default values, check log file at /var/log/foreman-installer/katello.log for more information
2021-04-13 01:16:25 [ERROR ] [root] RUBYLIB=/opt/theforeman/tfm/root/usr/share/gems/gems/kafo-6.3.0/lib/kafo/../..//modules: /opt/puppetlabs/bin/puppet apply /tmp/kafo_installation20210413-30980-kkg98t/dump_manifest.pp --config=/tmp/kafo_installation20210413-30980-kkg98t/puppet.conf 
2021-04-13 01:16:25 [ERROR ] [root] Warning: /etc/foreman-installer/custom-hiera.yaml: file does not contain a valid yaml hash
Error: Evaluation Error: Error while evaluating a Function Call, Lookup of key 'katello::params::enable_katello_agent' failed: Syntax error in string: %{alias('foreman_proxy_content::enable_katello_agent', 'undef', 'undef', 'false')} (file: /tmp/kafo_installation20210413-30980-kkg98t/dump_manifest.pp, line: 119, column: 346) on node pipe-katello-server-nightly-centos7.wareagle.example.com

2021-04-13 01:16:25 [ERROR ] [root] Could not get default values, cannot continue
[root@pipe-katello-server-nightly-centos7 kafo-6.3.0]# vim /usr/share/foreman-installer/config/foreman.hiera/scenario/katello.yaml
[root@pipe-katello-server-nightly-centos7 kafo-6.3.0]# foreman-installer --foreman-proxy-content-enable-katello-agent false
2021-04-13 01:16:36 [NOTICE] [root] Loading installer configuration. This will take some time.
Could not get default values, check log file at /var/log/foreman-installer/katello.log for more information
2021-04-13 01:16:38 [ERROR ] [root] RUBYLIB=/opt/theforeman/tfm/root/usr/share/gems/gems/kafo-6.3.0/lib/kafo/../..//modules: /opt/puppetlabs/bin/puppet apply /tmp/kafo_installation20210413-31254-zqrcev/dump_manifest.pp --config=/tmp/kafo_installation20210413-31254-zqrcev/puppet.conf 
2021-04-13 01:16:38 [ERROR ] [root] Warning: /etc/foreman-installer/custom-hiera.yaml: file does not contain a valid yaml hash
Error: Evaluation Error: Error while evaluating a Function Call, Lookup of key 'katello::params::enable_katello_agent' failed: Syntax error in string: %{lookup('foreman_proxy_content::enable_katello_agent', 'undef', 'undef', 'false')} (file: /tmp/kafo_installation20210413-31254-zqrcev/dump_manifest.pp, line: 119, column: 346) on node pipe-katello-server-nightly-centos7.wareagle.example.com

2021-04-13 01:16:38 [ERROR ] [root] Could not get default values, cannot continue
[root@pipe-katello-server-nightly-centos7 kafo-6.3.0]# vim /usr/share/foreman-installer/config/foreman.hiera/scenario/katello.yaml
[root@pipe-katello-server-nightly-centos7 kafo-6.3.0]# foreman-installer --foreman-proxy-content-enable-katello-agent false
2021-04-13 01:16:55 [NOTICE] [root] Loading installer configuration. This will take some time.
Could not get default values, check log file at /var/log/foreman-installer/katello.log for more information
2021-04-13 01:16:57 [ERROR ] [root] RUBYLIB=/opt/theforeman/tfm/root/usr/share/gems/gems/kafo-6.3.0/lib/kafo/../..//modules: /opt/puppetlabs/bin/puppet apply /tmp/kafo_installation20210413-31529-10cr9mo/dump_manifest.pp --config=/tmp/kafo_installation20210413-31529-10cr9mo/puppet.conf 
2021-04-13 01:16:57 [ERROR ] [root] Warning: /etc/foreman-installer/custom-hiera.yaml: file does not contain a valid yaml hash
Error: Evaluation Error: Error while evaluating a Function Call, Lookup of key 'katello::params::enable_katello_agent' failed: Syntax error in string: %{alias('foreman_proxy_content::enable_katello_agent', false)} (file: /tmp/kafo_installation20210413-31529-10cr9mo/dump_manifest.pp, line: 119, column: 346) on node pipe-katello-server-nightly-centos7.wareagle.example.com

2021-04-13 01:16:57 [ERROR ] [root] Could not get default values, cannot continue
[root@pipe-katello-server-nightly-centos7 kafo-6.3.0]# vim /usr/share/foreman-installer/config/foreman.hiera/scenario/katello.yaml
[root@pipe-katello-server-nightly-centos7 kafo-6.3.0]# foreman-installer --foreman-proxy-content-enable-katello-agent false
2021-04-13 01:17:19 [NOTICE] [root] Loading installer configuration. This will take some time.
Could not get default values, check log file at /var/log/foreman-installer/katello.log for more information
2021-04-13 01:17:22 [ERROR ] [root] RUBYLIB=/opt/theforeman/tfm/root/usr/share/gems/gems/kafo-6.3.0/lib/kafo/../..//modules: /opt/puppetlabs/bin/puppet apply /tmp/kafo_installation20210413-31804-1amle7g/dump_manifest.pp --config=/tmp/kafo_installation20210413-31804-1amle7g/puppet.conf 
2021-04-13 01:17:22 [ERROR ] [root] Warning: /etc/foreman-installer/custom-hiera.yaml: file does not contain a valid yaml hash
Error: Evaluation Error: Error while evaluating a Function Call, Lookup of key 'katello::params::enable_katello_agent' failed: Syntax error in string: %{alias('foreman_proxy_content::enable_katello_agent', :default_value => false)} (file: /tmp/kafo_installation20210413-31804-1amle7g/dump_manifest.pp, line: 119, column: 346) on node pipe-katello-server-nightly-centos7.wareagle.example.com
```

test/fixtures/manifests/basic.pp
```diff
@@ -40,9 +40,9 @@
   Boolean $remote = true,
   $server = 'mysql.example.com',
   $username = 'root',
-  Integer $pool_size = 10,
+  Integer $pool_size = $testing::params::pool_size,
   $file = undef,
-  $base_dir = undef) {
+  $base_dir = undef) inherits testing::params {
```
ekohl, line 45: This file is named `basic.pp`. Why do you modify this to be more complex?

ekohl (changes requested):
I'm not sure this is really where I want to go. Conceptually I think this is the wrong direction. In 3b154b57768545d0a5044df36c5b1ccefe53fa65 I explicitly chose to include Hiera so that we could override defaults from Puppet. Implementing this means we can no longer override defaults in our Hiera layer and rely on `--reset-PARAMETER`.

I think what we really found was that `alias` doesn't return `undef` if it's undefined. IMHO we should first figure out if that's correct. It feels like a bug to me.

ehelms:
> I think what we really found was that `alias` doesn't return `undef` if it's undefined. IMHO we should first figure out if that's correct. It feels like a bug to me.

Inside a yaml file for hiera, this is correct behavior:

> Note: A lookup resulting in an interpolation of `alias` referencing a non-existant key returns an empty string, not a Hiera "not found" condition.

Further, the versions of these functions and how they work inside the yaml are that they only accept a single argument and one cannot use the default value trick.

ekohl:
Then perhaps it's better to modify the code to deal with an empty string instead.

ehelms:
Closing in favor of updating the puppet code to handle this

PR #332, opened 2021-05-05
Fixes #32530: Enable passing a block to app_option for validations

ehelms:
The clamp API for options allows passing a block to perform validations
on the incoming value. This opens up this ability to options defined
within hooks.

### DIFF
--- a/lib/kafo/hook_context.rb
+++ b/lib/kafo/hook_context.rb
@@ -34,8 +34,8 @@ module Kafo
     #
     # @example
     #   app_option ['-n', '--noop'], :flag, 'Run puppet in noop mode?', :default => false
-    def app_option(*args)
-      self.kafo.class.app_option(*args)
+    def app_option(*args, &block)
+      self.kafo.class.app_option(*args, &block)
     end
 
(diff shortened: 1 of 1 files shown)


wbclark (commented):
I validated that clamp receives block arguments, and we already allow passing them in other parts of Kafo. I understand the basics of how block arguments work in general, but after reading https://projects.theforeman.org/issues/32530 it's still not clear to me how we might use them with an app_option in the hook context.

Therefore could you provide a simple example of how this would be used?

EDIT: In the time it took to write that, I see that you mentioned this from an installer PR. So I will read there, thanks.

ekohl (approved):
I wonder if we should also add `**kwargs` while we're at it. However, I think that's Ruby 2.5+ and Kafo is 2.4+ now. Let's get this in since this is a feature I've also missed in the past. In fact, I had a block validation in the installer before but found out it didn't work. There's more options we can validate this way.

ekohl:
> Therefore could you provide a simple example of how this would be used?

https://github.com/mdub/clamp#validation has more.

ehelms:
Here is a use in foreman-installer --
https://github.com/theforeman/foreman-installer/pull/679

PR #333, opened 2021-05-06
Fixes #32549: Implement multi hooks

ehelms:
Multi hooks are a special hook that allow you to define multiple hook
stages within a single file. This can to avoid duplication of things
like constants.

This takes over https://github.com/theforeman/kafo/pull/288 and add tests and documentation. Besides what I have added is there anything else you'd like to see?

### DIFF
--- a/test/fixtures/hooks/multi/10-print_hello_kafo.rb
+++ b/test/fixtures/hooks/multi/10-print_hello_kafo.rb
@@ -1,0 +1,13 @@
+boot do
+  app_option('--print-hello-kafo', :flag, 'Prints Hello Kafo when run is complete')
+end
+
+post('11-print_goodbye') do
+  puts "Goodbye"
+end
+
+post do
+  if app_value(:print_hello_kafo)
+    puts "Hello Kafo"
+  end
+end
--- a/README.md
+++ b/README.md
@@ -816,6 +816,32 @@ if app_value(:reset_foreman_db) && !app_value(:noop)
 end
 ```
 
+Hooks can additionally be defined by combining all related stages into a single file
+known as a Multi-stage hook. Multi-stage hooks live in a special directory inside
+the hooks directory: ```$installer_dir/hooks/multi```. Taking the previous example:
+
+```ruby
+# hooks/multi/10-reset_option_feature.rb
+boot do
+  app_option '--reset-foreman-db', :flag, 'Drop foreman database first? You will lose all data!', :default => false
+end
+
+pre do
+  if app_value(:reset_foreman_db) && !app_value(:noop)
+    `which foreman-rake > /dev/null 2>&1`
+    if $?.success?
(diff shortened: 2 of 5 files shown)


test/fixtures/hooks/multi/10-print_hello_kafo.rb
```diff
@@ -0,0 +1,9 @@
+boot do
```
ekohl, line 1: Other ideas:
* test that you can instantiate multiple hooks in the same file
* test that you can give hooks names which is important for ordering

ehelms: > * test that you can instantiate multiple hooks in the same file

Multiple hooks of the same *type* (e.g. 2 boot hooks) ? Since I technically am testing multiple hooks already :)

ehelms: > * test that you can instantiate multiple hooks in the same file
> * test that you can give hooks names which is important for ordering

I think I hit these in one go by adding an additional `post` hook with names and ordering expressed in the names and then testing the output is in the correct order.

ekohl: > Multiple hooks of the same _type_ (e.g. 2 boot hooks) ? Since I technically am testing multiple hooks already :)

Yes. Technically I think it's supported if you use names but it doesn't exist unless there's tests, right? :)

> I think I hit these in one go by adding an additional `post` hook with names and ordering expressed in the names and then testing the output is in the correct order.

Pretty sure I'd do the same thing.

test/fixtures/hooks/multi/10-print_hello_kafo.rb
```diff
@@ -0,0 +1,13 @@
+boot('10-print_hello_kafo') do
+  app_option('--print-hello-kafo', :flag, 'Prints Hello Kafo when run is complete')
+end
+
+post('11-print_goodbye') do
+  puts "Goodbye"
+end
+
+post('10-print_hello_kafo') do
```
ekohl, line 9: This now doesn't test that it inherits the default name. You can leave this out and it should inherit `10-print_hello_kafo` already
```suggestion
post do
```

test/fixtures/hooks/multi/10-print_hello_kafo.rb
```diff
@@ -0,0 +1,13 @@
+boot('10-print_hello_kafo') do
```
ekohl, line 1: Syntactically you're also allowed to write
```suggestion
boot '10-print_hello_kafo' do
```

Perhaps a bit of a style thing so I'll leave it up to you.

ehelms:
Updated to only name the one hook in the file to test correct ordering.

ekohl (approved):
Thanks for finishing this up.

PR #335, opened 2021-05-17
Drop workarounds for Ruby < 2.4

ehelms:


lib/kafo/kafo_configure.rb
```diff
@@ -525,8 +525,8 @@ def run_installation
               @progress_bar.update(line) if @progress_bar
             end
           rescue Errno::EIO # we reach end of input
-            exit_status = PTY.check(pid, true) if PTY.respond_to?(:check) # ruby >= 1.9.2
-            if exit_status.nil? # process is still running or we have old ruby so we don't know
+            exit_status = PTY.check(pid, true) if PTY.respond_to?(:check)
```
ekohl, line 528: Feels like you can just make it like this on Ruby 1.9.2+
```suggestion
            exit_status = PTY.check(pid, true)
```

ekohl (approved):
This could have been done as part of https://github.com/theforeman/kafo/commit/787085f6b10147dc85708b04134caed805cfd210 since it declared it as a Ruby 2.4+ project.

PR #336, opened 2021-05-17
Fixes #32611: Drop Puppet 5 support

ehelms:


ehelms:
Testing update needed prior to -- https://github.com/theforeman/jenkins-jobs/pull/65

modules/kafo_configure/manifests/init.pp
```diff
@@ -4,14 +4,7 @@
 #   $modulepath/config/answers.yaml
 #   /etc/kafo-configure/answers.yaml
 #
-# @param add_progress
-#   Whether to add a progress bar. Only works on Puppet < 6.
 class kafo_configure(
-  Boolean $add_progress = $::kafo_add_progress,
```
ekohl, line 10: If you're dropping this, please also drop all code for it:
https://github.com/theforeman/kafo/blob/a45b1ebedda66cdf460c3fe5bfbb3696f2bd198f/lib/kafo/puppet_command.rb#L76
https://github.com/theforeman/kafo/blob/a45b1ebedda66cdf460c3fe5bfbb3696f2bd198f/lib/kafo/puppet_command.rb#L82-L84
(and the tests)

ekohl (changes requested):
Please keep the README up to date (https://github.com/theforeman/kafo#requirements).

ekohl:
Also, I'd like to hold this off. Doing another major release is just a pain and we don't really gain a lot by doing so now, right?

ehelms:
> Also, I'd like to hold this off. Doing another major release is just a pain and we don't really gain a lot by doing so now, right?

If not now, then when? If we plan to drop this support, we benefit ourselves from doing it across the board and at the start of a release. The "major" bump aspect of it is pretty easy to orchestrate. What do you see as the pain ?

ekohl:
Cherry picks become more work.

ekohl:
In https://github.com/theforeman/kafo/pull/354 I went a step further and also cleaned up the progress bar support.

PR #338, opened 2021-08-13
Add new configuration option for puppet classes

ehelms:
This builds on the initial proposal from this RFC (https://community.theforeman.org/t/defaulting-puppet-to-off-in-the-katello-scenario/14553) and further thoughts within the RFC (https://community.theforeman.org/t/defaulting-puppet-to-off-in-the-katello-scenario/14553/8).

This approach is to apply the idea of keeping scenario configuration separated from the parameter value storage. The benefits and aim of this solution over a complete version 2 answer file is:

 1) This change is progressive and backwards compatible. Scenarios can add this new section to the configuration file and take advantage of the new features.
 2) This model allows adding new features over time as new keys can be added to the configuration hash (e.g. include params, controlling enablement state)
 3) By keeping this in the configuration file, the answers file can remain in standard hiera yaml format. This means no custom backend or manipulation of the parameters in the answers file to allow the hiera hierarchy to continue to work. 
 4) This change is much smaller than an over haul to add support for a V2 answer file would be and thus the risk is much lower to existing projects.

The initial features this adds are:

 1) Ability to specify if a puppet class should be presented to the user as something than can be disabled. If `can_disable` is set to false, the user is never presented with a configuration option to disable it.
 2) Ability to specify parameters that should not be presented to a user as a CLI option. Then the scenario can hide parameters that either clutter the help output or are not meant to be configured by a user in an installer setup.

Example: Set foreman to not be able to disable and exclude the version and db_manage parameter:

```
:classes:
  :foreman:
    :can_disable: false
    :exclude:
      - version
      - db-manage
```

** Why a dedicated `can_disable` flag instead of a multi-state `enabled` field (e.g. true/false/always) ?

Attempting to use a single field for enabled did not easily cover all the states we need. Those being:

 * Puppet class is enabled by default, but can be disabled
 * Puppet class is disabled by default, can be enabled and then later disabled
 * Puppet class is enabled by default but cannot be disabled
 * Puppet class is disabled by default, can be enabled but once enabled cannot be disabled

The last condition here not being covered by the original RFC proposal of using 'always'. This separate field let's us build the entire matrix of states and cleanly separates the state (enabled/disabled) and capability (can be disabled).

** Why not add an `include` field right now?

Since all parameters are included by default, we need only to currently specify which parameters to exclude.

### DIFF
--- a/lib/kafo/configuration.rb
+++ b/lib/kafo/configuration.rb
@@ -135,13 +135,21 @@ module Kafo
         @data.map do |name, values|
           if (class_config = app[:classes][name.to_sym])
             can_disable = class_config[:can_disable] || false
+            excluded_params = class_config[:exclude] || []
           else
             can_disable = true
+            excluded_params = []
           end
 
           enabled = !!values || values.is_a?(Hash)
 
-          puppet_mod = PuppetModule.new(name, configuration: self, enabled: enabled, can_disable: can_disable)
+          puppet_mod = PuppetModule.new(
+            name,
+            configuration: self,
+            enabled: enabled,
+            can_disable: can_disable,
+            excluded_params: excluded_params
+          )
           puppet_mod.parse
         end.sort
       end
--- a/lib/kafo/configuration.rb
+++ b/lib/kafo/configuration.rb
@@ -43,6 +43,7 @@ module Kafo
       ScenarioOption::KAFO_MODULES_DIR          => nil,
       ScenarioOption::CONFIG_HEADER_FILE        => nil,
       ScenarioOption::DONT_SAVE_ANSWERS         => nil,
+      ScenarioOption::CLASSES                   => {},
     }
 
     def self.get_scenario_id(filename)
@@ -132,8 +133,15 @@ module Kafo
         register_data_types
 
         @data.map do |name, values|
+          if (class_config = app[:classes][name.to_sym])
+            can_disable = class_config[:can_disable] || false
+          else
+            can_disable = true
+          end
+
           enabled = !!values || values.is_a?(Hash)
-          puppet_mod = PuppetModule.new(name, configuration: self, enabled: enabled)
+
+          puppet_mod = PuppetModule.new(name, configuration: self, enabled: enabled, can_disable: can_disable)
           puppet_mod.parse
         end.sort
       end
--- a/lib/kafo/configuration.rb
+++ b/lib/kafo/configuration.rb
@@ -130,7 +130,12 @@ module Kafo
     def modules
       @modules ||= begin
         register_data_types
-        @data.keys.map { |mod| PuppetModule.new(mod, configuration: self).parse }.sort
+
+        @data.map do |name, values|
+          enabled = !!values || values.is_a?(Hash)
+          puppet_mod = PuppetModule.new(name, configuration: self, enabled: enabled)
+          puppet_mod.parse
+        end.sort
       end
     end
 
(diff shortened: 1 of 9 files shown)


lib/kafo/configuration.rb
```diff
@@ -43,6 +43,7 @@ class Configuration
       ScenarioOption::KAFO_MODULES_DIR          => nil,
       ScenarioOption::CONFIG_HEADER_FILE        => nil,
       ScenarioOption::DONT_SAVE_ANSWERS         => nil,
+      ScenarioOption::CLASSES                   => {},
```
ehelms, line 46: The standard nomenclature we have been using is `modules`; but what we really seem to specify everywhere are puppet classes not puppet modules. Thus I decided to aim for using the correct representation.

test/acceptance/kafo_configure_test.rb
```diff
@@ -340,5 +340,66 @@ module Kafo
...
+
+        answers = {'testing' => false}
+        File.open(KAFO_ANSWERS, 'w') do |file|
+          file.write(answers.to_yaml)
+        end
+
+        code, out, err = run_command '../bin/kafo-configure --help'
+        _(code).must_equal 0, err
+        _(out).must_include "--enable-testing"
+      end
```
evgeni, line 378: does that mean that when I've run `that-installer --enable-testing` once, the next run will error out because `--enable-testing` is not an option anymore? I can see some (a|se)nsible users cry about that.

lib/kafo/kafo_configure.rb
```diff
@@ -388,11 +388,15 @@ def set_options
       end
 
       modules.each do |mod|
-        app_option d("--[no-]enable-#{mod.name}"), :flag, "Enable '#{mod.name}' puppet module",
-                   :default => mod.enabled?
+        if mod.can_disable?
+          app_option d("--[no-]enable-#{mod.name}"), :flag, "Enable '#{mod.name}' puppet module (current: #{mod.enabled?})", :default => mod.enabled?
+        elsif !mod.enabled?
+          app_option d("--enable-#{mod.name}"), :flag, "Enable '#{mod.name}' puppet module (current: #{mod.enabled?})"
+        end
```
evgeni, line 395: I guess https://github.com/theforeman/kafo/pull/338/files#r701787011 should rather have been here, but :woman_shrugging:

test/acceptance/kafo_configure_test.rb
ehelms: That is a fair point. My thought was, if the option does nothing, why include it. There are a couple ideas:

 1) If a module is already enabled and cannot be disabled, promote it to `--full-help` to get it out of the users way but still exist
 2) I have been thinking of a better option for ansible-style users, where users can provide a file as input that maps to installer parameters. That would be a nicer interface for those type of users. Though it may suffer from the same problem (no such option).

evgeni: Demoting it to `full-help` sounds good to me.

While we can (and should) make the Ansible life easier by making it *understand* the installer, we also have users out there who have a 50+ page PDF "how to install/upgrade $project" and they will re-use the same installer parameters on every installer invocation.

Is that (repetition of the params on every invocation) needed? Of course not.
Will the users be totally confused that they have feature X enabled, and yet the installer tells them there is no such thing as feature X (which is how they will read the absence of `--enable-X`)? Absolutely.
(As me how I know.)

PR #340, opened 2021-08-20
Cleanup configuration file

ehelms:


test/kafo/configuration_test.rb
```diff
@@ -4,8 +4,8 @@
 module Kafo
   describe Configuration do
     let(:basic_config_file) { ConfigFileFactory.build('basic', BASIC_CONFIGURATION).path }
-    let(:basic_config) { Kafo::Configuration.new(basic_config_file, false) }
-    let(:old_config) { Kafo::Configuration.new(basic_config_file, false) }
+    let(:basic_config) { Kafo::Configuration.new(basic_config_file, Kafo::KafoConfigure.logger, false) }
```
ekohl, line 7: Perhaps a bit shorter in most places:
```suggestion
    let(:logger) { Kafo::KafoConfigure.logger }
    let(:basic_config) { Kafo::Configuration.new(basic_config_file, logger, false) }
```

lib/kafo/configuration.rb
```diff
@@ -49,11 +49,11 @@ def self.get_scenario_id(filename)
       File.basename(filename, '.yaml')
     end
 
-    def initialize(file, persist = true)
+    def initialize(file, logger, persist = true)
```
ekohl, line 52: How about keeping the API compatible:
```suggestion
    def initialize(file, persist = true, logger = KafoConfigure.logger)
```

I'm also not sure if we can rely on this (I think it may be Ruby 2.5+), but you can also enforce a named argument:
```suggestion
    def initialize(file, persist = true, logger:)
```

Or with a default:
```suggestion
    def initialize(file, persist = true, logger: KafoConfigure.logger)
```

Then you must call it as:
```ruby
Configuration.new(config, logger: KafoConfigure.logger)
```

lib/kafo/configuration.rb
```diff
@@ -135,7 +134,7 @@ def modules
     end
 
     def module(name)
-      modules.find { |m| m.name == name }
+      modules.find { |m| m.identifier == name }
```
ekohl, line 137: This looks like a bugfix. Could you provide a bit more info about why it's wrong?

ekohl (commented):
Overall I think this makes sense.

lib/kafo/configuration.rb
ehelms: I tried to update the commit message with more information. The most telling is to look at the hook context documentation. Unfortunately, I think `name` in some areas of code refers to the Puppet naming scheme with `::` and in other places it references this internal "name" construct that Kafo creates -- https://github.com/theforeman/kafo/blob/master/lib/kafo/puppet_module.rb#L175

PR #341, opened 2021-08-26
Ensure buffered log messages contain the proper timestamp

ehelms:
Before logging is configured, log messages are stored in an internal
buffer to be dumped at the point at which logging is configured. At
the time the message was being dumped the timestamp was being calculated
making it appear as if a set of log messages all happened at the same
time. This gave no sense that those initial log messages and operations
might be taking long time in some cases.

This moves to storing those buffered messages as LogEvent's that
capture the time of creation such that when those messages are then
dumped to the logs they contain the timestamp of when the message
was actually created rather than the timestamp of when the message
was dumped.

### DIFF
--- a/lib/kafo/logger.rb
+++ b/lib/kafo/logger.rb
@@ -13,7 +13,13 @@ module Kafo
 
     def log(level, *args, &block)
       if Logging.buffering?
-        Logging.to_buffer(@name, level, args, &block)
+        if block_given?
+          data = yield
+        else
+          data = args
+        end
+
+        Logging.to_buffer(@name, ::Logging::LogEvent.new(@name, ::Logging::LEVELS[level.to_s], data, false))
       else
         Logging.dump_buffer if Logging.dump_needed?
         @logger.send(level, *args, &block)
--- a/lib/kafo/logging.rb
+++ b/lib/kafo/logging.rb
@@ -118,7 +118,7 @@ module Kafo
 
       def dump_buffer
         @buffer.each do |log|
-          ::Logging.logger[log[0]].send(log[1], *([log[2]].flatten(2)), &log[3])
+          ::Logging.logger[log[0]].send(:log_event, log[1])
         end
         @buffer.clear
       end


wbclark:
I noticed almost duplicate lines here:

```
2021-08-26 18:15:50 [INFO  ] [boot] All hooks in group boot finished
2021-08-26 18:15:50 [INFO  ] [root] Boot hooks done
2021-08-26 18:15:50 [NOTICE] [root] Loading installer configuration. This will take some time.
```

That were absent when I tested on my Katello 4.2:

```
2021-08-27 23:04:38 [INFO  ] [boot] All hooks in group boot finished
2021-08-27 23:04:38 [NOTICE] [root] Loading installer configuration. This will take some time.
```

I think it's unrelated to this PR but worth pointing out.... if there is some utility in having both of those messages logged, one from the boot hooks logger and one from the root logger, then I think at least one of them should be at level `DEBUG` instead of `INFO`.

Also unrelated to the PR, but interesting to me regardless: boot hooks in your case took over 2 minutes:

```
2021-08-26 18:13:18 [INFO  ] [boot] Executing hooks in group boot
2021-08-26 18:15:50 [INFO  ] [boot] All hooks in group boot finished
```

While on my environment they completed in about 2 seconds:

```
2021-08-27 23:04:36 [INFO  ] [boot] Executing hooks in group boot
2021-08-27 23:04:38 [INFO  ] [boot] All hooks in group boot finished
```

I can confirm that with this change applied, the difference in timestamps is seen in log messages from the buffer before configuration is loaded (they were not before).

lib/kafo/logger.rb
```diff
@@ -13,7 +13,7 @@ def initialize(name = 'root')
 
     def log(level, *args, &block)
       if Logging.buffering?
-        Logging.to_buffer(@name, level, args, &block)
+        Logging.to_buffer(@name, ::Logging::LogEvent.new(@name, ::Logging::LEVELS[level.to_s], *args, false))
```
wbclark, line 16: I noticed you dropped the `&block` here, and from what I can tell neither `Logging.to_buffer()` nor `::Logging::LogEvent.new()` has anything to do with such an argument.

It looks like `&block` _can_ be passed to `::Logging::Logger.some_level` ( https://github.com/TwP/logging/blob/18cc826ae65fee63c120a297d3b0235337980058/lib/logging/logger.rb#L112 ) which we support in theory with L19: `@logger.send(level, *args, &block)`

If I understood that correctly it means we have the capability to do things like:

```ruby
logger.notice do
  a = something()
  b = something_else()
  even_more_stuff(a, b)
end
```

which would log an event at level `NOTICE` with `data = even_more_stuff(a, b)`

That's pretty neat, but I looked for places where we actually use it and I couldn't find any

```shell
$ egrep -nr 'log\(|log |logger\.(debug|info|notice|warn|error)' lib/ bin/ test/kafo/logger_test.rb | egrep ' {|&.+\)| do'
lib/kafo/logger.rb:14:    def log(level, *args, &block)
lib/kafo/logger.rb:25:        log(level, *args, &block)
test/kafo/logger_test.rb:23:      it "add log messages to buffer" do
$
```

wbclark: I think IF we had tried to use that pattern `logger.some_level { some_block }` when `logger.buffering?` then before this change `logger.to_buffer` just drops the `&block` anyway, and the same would happen now with the way you have it written.

I think that's fine perhaps -- we already don't actually use this as far as I could tell -- but in that case maybe we should make things simple and drop all this `&block` stuff on L14, L19, L24, and L25 since it would already fail silently some of the time.

wbclark: Or to keep it around we could try to make `Logging::to_buffer` do something with `&block`. Something like...

```ruby
def to_buffer(*args, &block)
  return buffer << yield if block_given?
  buffer << args
end
```
and,

```suggestion
        Logging.to_buffer(@name, ::Logging::LogEvent.new(@name, ::Logging::LEVELS[level.to_s], *args, false), &block)
```

then you could do stuff like (totally made up example to illustrate the idea):

```ruby
basic_checks.each do |check|
  logger.notice(name = check.name) do
    result = "success!" if check.run() rescue "FAIL!"
    "ran basic check: #{name} with result: #{result}"
  end
end
```

no idea if we'd ever have a reason to use something like this where we pass blocks to `logger.level` but I guess we should either support it always, including when `logger.buffering?`, or else not at all.

ehelms:
The logs are from a local environment where there may be differen tlog messages than what you experience, and the boot hook is longer in mine because I was playing around with parser caches. I would not judge this change on the *actual* messages in my example, but the intent of how those log messages are displayed to the user with respect to timestamp.

ehelms:
I have added back the block support

wbclark:
> I have added back the block support

Oh, that's a nice way of doing it

wbclark (approved):
LGTM

PR #346, opened 2022-02-11
Add parameter to pass in a default for custom config

ekohl:
This is a potential solution for a problem in https://github.com/theforeman/foreman-installer/pull/744

### DIFF
--- a/lib/kafo/configuration.rb
+++ b/lib/kafo/configuration.rb
@@ -107,8 +107,8 @@ module Kafo
       end
     end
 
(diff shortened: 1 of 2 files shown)


evgeni:
Regardless that I would like not to need this for https://github.com/theforeman/foreman-installer/pull/744, I think this is a good idea and we should get that.

Maybe with a test? (says the person who had the pleasure to distill the meaning of default values in such helpers in foreman core) ;)

PR #347, opened 2022-04-23
Add migration_name and named logger to MigrationContext

wbclark:


ehelms:
If you have before and after output that would help in reviewing.

PR #348, opened 2022-04-23
Add method MigrationContext.keep_migration

wbclark:
Such a migration may alter configuration and answers, or it may noop, but is not required to be recorded in applied migrations and may run more than once, until some condition is met. One simple application would be a backwards compatible migration that noops until a certain target version is reached.

Work in progress, builds on https://github.com/theforeman/kafo/pull/347


ekohl:
I'm not sure I understand the use case for this. Hooks are already executed every time and mixing migrations in this way is going to be confusing. For now I'm closing this, but if you have a good use case then we can reopen it.

PR #354, opened 2023-05-05
Fixes #32611: Drop Puppet < 7 support & require Ruby 2.7+

ekohl:
Also drops progress bar support which removes a dependency.


evgeni:
you probably want https://github.com/theforeman/jenkins-jobs/pull/335 so that CI can be green :)

ekohl:
[test kafo]

evgeni:
```
kafo-configure::default args.test_0002_must fail if validations fail
```

🤔

lib/kafo/kafo_configure.rb
```diff
@@ -473,8 +467,10 @@ def validate_all(logging = true)
       logger.info "Running validation checks."
       results = enabled_params.map do |param|
         result = param.valid?
-        errors = param.validation_errors.join(', ')
-        progress_log(:error, "Parameter #{with_prefix(param)} invalid: #{errors}", logger) if logging && !result
+        if logging && !result
+          errors = param.validation_errors.join(', ')
+          logger.error("Parameter #{with_prefix(param)} invalid: #{errors}")
+        end
```
ekohl, line 473: I wonder why this is different and now fails the tests.

evgeni: probably because it (now) writes to stderr, but our tests only capture stdout?

evgeni: (or the other way round)

ekohl: That's likely, though I wonder how. Right now I'm fighting to get the test suite to work on my desktop with Ruby 3.2.

evgeni: adding an explicit `$stderr.puts "Parameter #{with_prefix(param)} invalid: #{errors}"` here (obviously) helps, and stdout doesn't contain any of this. which makes me think `logger.error` does end up somewhere, but not on the terminal.

ekohl: I removed this part:

https://github.com/theforeman/kafo/blob/a24cec27eaf10ac6deac635e900bcebf1539c437/lib/kafo/kafo_configure.rb#L549-L552

And replaced it what I thought was equivalent.

But why did it end up on stderr before? I think we wouldn't have a progress bar, but we sort of did?

ekohl: No, I think it's with `--no-verbose`, but `--verbose` is also the default I think. So I think that's it: powerbar still printed the error to stderr for us. I'll have to rethink this PR.

evgeni: Indeed, when I change:
```diff
-        code, _, err = run_command '../bin/kafo-configure --testing-pool-size=fail'
+        code, _, err = run_command '../bin/kafo-configure --testing-pool-size=fail --verbose'
```
the test fails on master too ;)

evgeni:
Meh, we missed that https://github.com/theforeman/kafo#requirements also mentions Puppet versions

ekohl:
I looked through the README. Not sure how I missed that one.

ekohl:
https://github.com/theforeman/kafo/pull/356

PR #358, opened 2023-07-25
Add CI workflow using GHA

archanaserver:
This pull request adds a GitHub Actions CI workflow to this project.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,46 @@
+name: CI
+
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+
+concurrency:
+  group: ${{ github.ref_name }}
+  cancel-in-progress: true
+
+jobs:
+    
+  rubocop:
+    name: Rubocop
+    uses: theforeman/actions/.github/workflows/rubocop.yml@v0
+
+  test:
+    name: Tests
+    needs: rubocop
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        ruby:
+          - "2.7"
+        puppet:
+          - "7"
+    
+    env:
+      PUPPET_VERSION: ${{ matrix.puppet }}
+      
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v3
+
+      - name: Set up Ruby environment
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ${{ matrix.ruby }}
+          bundler-cache: true
+
+      - name: Run tests
+        run: |
+          bundle exec rake test test:acceptance
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -139,7 +139,7 @@ Style/SlicingWithRange:
   Enabled: true
 
 Lint/MixedRegexpCaptureTypes:
-  Enabled: false
+  Enabled: true
 
 Lint/ShadowingOuterLocalVariable:
   Enabled: true
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -3,10 +3,11 @@ inherit_from: .rubocop_todo.yml
 AllCops:
   TargetRubyVersion: '2.7'
   Exclude:
-    - "Gemfile"
-    - "Rakefile"
-    - "test/**/*"
-    - "modules/**/*"
+    - 'Gemfile'
+    - 'Rakefile'
+    - 'test/**/*'
+    - 'modules/**/*'
+    - 'vendor/**/*'
 
 Bundler/OrderedGems:
   Enabled: false
@@ -138,7 +139,7 @@ Style/SlicingWithRange:
   Enabled: true
 
 Lint/MixedRegexpCaptureTypes:
-  Enabled: true
+  Enabled: false
 
 Lint/ShadowingOuterLocalVariable:
   Enabled: true
--- a/Rakefile
+++ b/Rakefile
@@ -31,11 +31,12 @@ namespace 'test' do
   end
 end
 
-begin
-  require 'rubocop/rake_task'
-  RuboCop::RakeTask.new
-rescue LoadError
-  puts 'Rubocop not loaded'
+require 'rubocop/rake_task'
+RuboCop::RakeTask.new(:rubocop) do |task|
+  # These make the rubocop experience maybe slightly less terrible
+  task.options = ['--display-cop-names', '--display-style-guide', '--extra-details']
+  # Use Rubocop's Github Actions formatter if possible
+  task.formatters << 'github' if ENV['GITHUB_ACTIONS'] == 'true'
 end
 
 CLEAN.include 'test/tmp'
--- a/lib/kafo/kafo_configure.rb
+++ b/lib/kafo/kafo_configure.rb
@@ -519,7 +519,7 @@ module Kafo
             method, message = log_parser.parse(line)
             progress_log(method, message, logger)
 
-            if (output = line.match(/(?:.+\]): Starting to evaluate the resource( \((?<count>\d+) of (?<total>\d+)\))?/))
+            if (output = line.match(/(?:.+\]): Starting to evaluate the resource(?: \((?<count>\d+) of (?<total>\d+)\))?/))
               if (output[:count].to_i % 250) == 1 && output[:count].to_i != 1
                 logger.notice("#{output[:count].to_i - 1} configuration steps out of #{output[:total]} steps complete.")
               end
--- a/test/kafo/execution_environment_test.rb
+++ b/test/kafo/execution_environment_test.rb
@@ -54,10 +54,10 @@ module Kafo
         # Create files with different timestamps
         oldest = File.join(hostname_dir, 'oldest.yaml')
         FileUtils.touch(oldest)
-        sleep(0.001)
+        sleep(0.01)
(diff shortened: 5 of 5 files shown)


ehelms:
I assume this would aim to replace the Jenkins job testing which I am good with to make this git repo more light weight. @ekohl  thoughts?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,14 @@
+name: CI
+
+on: pull_request
+
+concurrency:
+  group: ${{ github.ref_name }}
+  cancel-in-progress: true
+
+jobs:
```
ekohl, line 9: I'd also run RuboCop
```suggestion
jobs:
  rubocop:
    name: Rubocop
    uses: theforeman/actions/.github/workflows/rubocop.yml@v0

```

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,14 @@
...
+  test:
+    name: Tests
+    uses: theforeman/actions/.github/workflows/test-gem.yml@v0
+    with:
+      command: bundle exec rake test
```
ekohl, line 14: And also run acceptance tests
```suggestion
      command: bundle exec rake test test:acceptance
```

ekohl (changes requested):
Overall I can see the benefits, but running RuboCop &  the acceptance tests is important.

I'd also enhance `Rakefile` with something similar as:
https://github.com/voxpupuli/voxpupuli-rubocop/blob/master/lib/voxpupuli/rubocop/rake.rb
In particular the formatters bit helps a lot because then you get annotations in your PRs on the lines that have errors, which is very useful.

ekohl:
Oh, and I also forgot that you want a test matrix which takes Puppet versions into account, which it currently doesn't. Kafo is much more complex in that sense and probably means you can't use the reusable action.

archanaserver:
> Oh, and I also forgot that you want a test matrix which takes Puppet versions into account, which it currently doesn't. Kafo is much more complex in that sense and probably means you can't use the reusable action.

@ekohl can you help me to understand it more, i can't get it!

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,56 @@
...
+        puppet: ["7"]
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v2
+
+      - name: Set up Puppet environment
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y puppet-agent=${{ matrix.puppet }}
+
```
ekohl, line 47: Please use `PUPPET_VERSION`, see https://github.com/theforeman/kafo/blob/4d3440c02a6592a3de34e8205688a118aa5cf360/Gemfile#L8

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,56 @@
...
+    name: Tests
+    runs-on: ubuntu-20.04
+    strategy:
+      fail-fast: false
+      matrix:
+        ruby: ["2.7"]
+        puppet: ["7"]
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v2
```
ekohl, line 41:
```suggestion
        uses: actions/checkout@v3
```

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,56 @@
...
+          bundle exec rubocop
+
+  test:
+    name: Tests
+    runs-on: ubuntu-20.04
+    strategy:
+      fail-fast: false
+      matrix:
+        ruby: ["2.7"]
+        puppet: ["7"]
```
ekohl, line 38:
```suggestion
        ruby:
          - "2.7"
        puppet:
          - "7"
```

.github/workflows/ci.yaml
ekohl: And you can set it like this: https://github.com/voxpupuli/beaker-hostgenerator/blob/61c7244c239c5e23a85ea3acb4952ff9d0ea9630/.github/workflows/test.yml#L35-L37

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,48 @@
...
+    name: Tests
+    runs-on: ubuntu-20.04
+    strategy:
+      fail-fast: false
+      matrix:
+        ruby:
+          - "2.7"
+    
+    env:
+      PUPPET_VERSION: "7"
```
ekohl, line 29: Right now we only support Puppet 7, but we want to support Puppet 8 as well. Please include the Puppet version in the test matrix.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,48 @@
...
+      
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v3
+
+      - name: Set up Puppet environment
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y "puppet-agent=${{ env.PUPPET_VERSION }}""
```
ekohl, line 38: Installation via apt shouldn't be needed since it's pulled in as a gem.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,48 @@
...
+
+jobs:
+    
+  rubocop:
+    name: Rubocop
+    uses: theforeman/actions/.github/workflows/rubocop.yml@v0
+
+  test:
+    name: Tests
+    runs-on: ubuntu-20.04
```
ekohl, line 21: Any particular reason you're using `ubuntu-20.04` instead of `ubuntu-latest`?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,48 @@
...
+  cancel-in-progress: true
+
+jobs:
+    
+  rubocop:
+    name: Rubocop
+    uses: theforeman/actions/.github/workflows/rubocop.yml@v0
+
+  test:
+    name: Tests
```
ekohl, line 20: The tests in Kafo are quite slow (especially the acceptance tests). Here I'd make the tests depend on RuboCop. That way you avoid running them if there are simple mistakes.
```suggestion
    name: Tests
    needs: rubocop
```

For gems where the tests are very fast I generally don't do that, but here it's worth it.

ekohl (changes requested):

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,47 @@
...
+    needs: rubocop
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        ruby:
+          - "2.7"
+        puppet:
+          - "7"
+          - "8"
```
ekohl, line 30: I don't think Puppet 8 & Ruby 2.7 is a valid combination. For now I'd leave it out. I was only suggesting to make sure the rest of the CI workflow is ready for it.
```suggestion
```

ekohl (changes requested):
Somehow RuboCop is picking up items from the vendor directory. I think you need to add `vendor/**/*` to `Exclude` in `.rubocop.yml`.

.github/workflows/ci.yaml
archanaserver: @ekohl oh i missed it, you were talking about matrix combination in this way? (i guess we can keep it, then!)
  ```
    matrix:
        ruby:
          - "2.7"
          - "3.1"
        puppet:
          - "7"
          - "8"
        include:
          - ruby: "2.7"
            puppet: "7"
          - ruby: "3.1"
            puppet: "8"
```

also i was looking for that puppet doc where i can see the compatible ruby version with the puppet version, can you help me with it?

Comment (ekohl) [reply .github/workflows/ci.yaml:30]: For this specific PR I would keep Puppet 8 and Ruby 3 out of consideration. Limit it to porting over what's already in Jenkins. There's a good chance you find more when you start adding Ruby 3 & Puppet 8 tests. Getting this in is a better first step. Otherwise you risk the PR dragging on for a long time.

Comment (archanaserver) [reply .github/workflows/ci.yaml:30]: okay ack! thanks for the info.

Comment (archanaserver) [reply .github/workflows/ci.yaml:30]: okay ack! thanks for the info.

File: .rubocop.yml [hunk 11]
```diff
@@ -138,7 +139,7 @@ Style/SlicingWithRange:
   Enabled: true
 
 Lint/MixedRegexpCaptureTypes:
-  Enabled: true
+  Enabled: false
```
ekohl, line 142: Can you test if it still passes with this enabled?
```suggestion
  Enabled: true
```

archanaserver: @ekohl it is enabled on my previous changes https://github.com/theforeman/kafo/commit/80f52596a07bddaecb2f0233072baeb1b081056d but now it is asking me to replace the named captures with numbered captures to avoid mixing them and which I did previously and that didn't look good! And the previous change is not satisfying this cop now. Do you have any suggestions to fix this or resources to look into, i think i'm missing something here.

ekohl: We missed a group:
https://github.com/theforeman/kafo/blob/a1b1b12d47f2e0dd26629452809c915e40a861b4/lib/kafo/kafo_configure.rb#L522

This needs to be:
```ruby
            if (output = line.match(/(?:.+\]): Starting to evaluate the resource(?: \((?<count>\d+) of (?<total>\d+)\))?/))
```

archanaserver: Thank @ekohl, updated the changes.

ekohl (approved):

PR #359, opened 2023-08-07
Fix some Rubocop offenses

archanaserver:
https://github.com/theforeman/kafo/pull/358


ekohl (approved):

ekohl (approved):
After rebasing on https://github.com/theforeman/kafo/pull/360 I think it should pass.

PR #361, opened 2023-08-10
Fix rubocop offenses

archanaserver:


.rubocop.yml
```diff
@@ -64,7 +64,7 @@ Style/HashSyntax:
   Enabled: false
 
 Style/HashTransformKeys:
-  Enabled: true
+  Enabled: false
```
ekohl, line 67: I'd keep this enabled.

ekohl (changes requested):

lib/kafo/kafo_configure.rb
```diff
@@ -523,7 +523,7 @@ def run_installation
               method, message = log_parser.parse(line)
               progress_log(method, message, logger)
 
-              if (output = line.match(/(.+\]): Starting to evaluate the resource( \((?<count>\d+) of (?<total>\d+)\))?/))
+              if (output = line.match(/(.+\]): Starting to evaluate the resource( \((\d+) of (\d+)\))?/))
```
ekohl, line 526: What you're removing are called named captures and are used on the line below. https://www.leighhalliday.com/named-captures-ruby-regular-expressions explains them. They are extremely useful in complex regular expressions because the numbers are very easy to mess up.

https://docs.rubocop.org/rubocop/cops_lint.html#lintmixedregexpcapturetypes explains the cop. The way to deal with it is to use `?:` for groups you discard
```suggestion
              if (output = line.match(/(?:.+\]): Starting to evaluate the resource( \((?<count>\d+) of (?<total>\d+)\))?/))
```

ekohl (changes requested):

.rubocop_todo.yml
```diff
@@ -602,15 +602,12 @@ Style/RaiseArgs:
...
-    - 'lib/kafo/data_types/not_undef.rb'
-    - 'lib/kafo/data_types/optional.rb'
+  Enabled: false
 
 # Offense count: 1
 # Cop supports --auto-correct.
 Style/RedundantParentheses:
-  Exclude:
-    - 'lib/kafo/logging.rb'
+  Enabled: false
```
ekohl, line 610: Perhaps fix these 2 as well while you're at it?

.rubocop.yml
```diff
@@ -132,3 +132,27 @@ Layout/SpaceInsideHashLiteralBraces:
...
+  Enabled: true
+
+Style/SymbolProc:
+  Enabled: true
+
+Style/RegexpLiteral:
+  Enabled: true
+
+Style/RedundantRegexpEscape:
+  Enabled: true
```
ekohl, line 158: Can you ensure there's a line ending? Most editors can be configured to do so. https://medium.com/@alexey.inkin/how-to-force-newline-at-end-of-files-and-why-you-should-do-it-fdf76d1d090e and https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline have different reasons why it's good.

ekohl (changes requested):

lib/kafo/puppet_command.rb
```diff
@@ -43,15 +43,15 @@ def self.search_puppet_path(bin_name)
       File.join([bin_path, bin_name].compact)
     end
 
-    def self.is_aio_puppet?
+    def self.aio_puppet?
```
ekohl, line 46: You're changing this method name, but not the tests for it. That causes test failures. Please also change it in `test/kafo/puppet_command_test.rb`.

archanaserver: This one is fixed now!

archanaserver: @ekohl is this ready to get in? so I can rebase https://github.com/theforeman/kafo/pull/358 on top of this one, and make it ready.

ekohl (approved):

PR #362, opened 2023-08-24
Store puppet_report as an instance variable

ekohl:
Somehow it wasn't working and storing it as an instance variable is cleaner anyway.

Found via https://github.com/theforeman/foreman-installer/pull/878

Fixes: 0612410091e7c74f8493fbaec1df9a516451018a ("Make Puppet reports available from the ExecEnv")


ekohl:
And that's why it was still a draft, didn't have time to fix tests before I had to run out. Should be fixed now.

PR #367, opened 2023-11-10
Fixes #36912 - support multi-line type definitions

evgeni:
This strips line breaks from type definitions, thus making them parseable again.

This will probably horribly break if type manifests would contain anything else but type definitions, but Puppet recommends [1] to store them in separate files and only have one type alias per file.

[1] https://www.puppet.com/docs/puppet/7/lang_type_aliases

### DIFF
--- a/lib/kafo/data_type_parser.rb
+++ b/lib/kafo/data_type_parser.rb
@@ -8,9 +8,26 @@ module Kafo
 
     def initialize(manifest)
       @logger = KafoConfigure.logger
-      @types = {}
+
+      lines = []
+      type_line_without_newlines = +''
       manifest.each_line do |line|
-        if (type = TYPE_DEFINITION.match(line.force_encoding("UTF-8")))
+        line = line.force_encoding("UTF-8").strip
+        next if line.start_with?('#')
+
+        line = line.split(' #').first.strip
+        if line =~ TYPE_DEFINITION
+          lines << type_line_without_newlines
+          type_line_without_newlines = line
+        else
+          type_line_without_newlines << line
+        end
+      end
+      lines << type_line_without_newlines
+
+      @types = {}
+      lines.each do |line|
+        if (type = TYPE_DEFINITION.match(line))
           @types[type[1]] = type[2]
         end
       end
--- a/test/kafo/data_type_parser_test.rb
+++ b/test/kafo/data_type_parser_test.rb
@@ -24,6 +24,30 @@ module Kafo
       it { _(parser.types).must_equal({'Ipv4' => 'Pattern[/^(\d+)\.(\d+)\.(\d+)\.(\d+)$/]'}) }
     end
 
+    describe "parse pattern with a hash inside" do
+      let (:file) do
+        <<~'PUPPET'
+        type Stdlib::Email = Pattern[/\A[a-zA-Z0-9.!#$%&'*+\/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*\z/]
+        PUPPET
+      end
+      it { _(parser.types).must_equal({"Stdlib::Email"=>"Pattern[/\\A[a-zA-Z0-9.!\#$%&'*+\\/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*\\z/]"}) }
+    end
+
+    describe "parse multiline alias" do
+      let(:file) { "type IP = Variant[\n  IPv4,\n  IPv6,\n]" }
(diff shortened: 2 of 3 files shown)


lib/kafo/data_type_parser.rb
```diff
@@ -8,9 +8,26 @@ class DataTypeParser
...
-      @types = {}
+
+      lines = []
+      last_line = ''
       manifest.each_line do |line|
-        if (type = TYPE_DEFINITION.match(line.force_encoding("UTF-8")))
+        line = line.force_encoding("UTF-8").strip
+        next if line.start_with?('#')
+
+        line = line.split('#').first.strip
```
ekohl, line 18: This will break if you use a Pattern with `#` in it.

test/kafo/data_type_parser_test.rb
```diff
@@ -24,6 +24,16 @@ module Kafo
       it { _(parser.types).must_equal({'Ipv4' => 'Pattern[/^(\d+)\.(\d+)\.(\d+)\.(\d+)$/]'}) }
     end
 
+    describe "parse multiline alias" do
+      let(:file) { "type IP = Variant[\n  IPv4,\n  IPv6,\n]" }
+      it { _(parser.types).must_equal({'IP' => 'Variant[IPv4,IPv6,]'}) }
```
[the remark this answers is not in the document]
evgeni: added a test to ensure that it's fine

ekohl (commented):
What if we use puppet-strings to parse it? There's this part in the JSON:
```json
  "data_type_aliases": [
    {
      "name": "Candlepin::LogLevel",
      "file": "types/loglevel.pp",
      "line": 2,
      "docstring": {
        "text": "",
        "tags": [
          {
            "tag_name": "summary",
            "text": "A log4j log level"
          }
        ]
      },
      "alias_of": "Enum['ALL', 'DEBUG', 'INFO', 'WARN', 'ERROR', 'FATAL', 'OFF', 'TRACE']"
    }
  ],
```

In https://github.com/theforeman/kafo/pull/343 I started to utilize the bulk parsing approach. Current kafo_parsers calls `puppet strings` for every file. The bulk parsing approach is to call it for multiple files (or even for a whole module at once) and then extract the data from that file.

Since the data type aliases are also in there, we could leverage that and avoid writing our own parser.

lib/kafo/data_type_parser.rb
evgeni: So will the current regex, no?

https://github.com/theforeman/kafo/blob/4b44b57ca54f3fb6cc1e46ed511093a6dd61f519/lib/kafo/data_type_parser.rb#L5

evgeni:
> What if we use puppet-strings to parse it?

This sounds like the correct approach long term, but it also seems like a bigger re-work is required to get that going correctly?

lib/kafo/data_type_parser.rb
evgeni: However, https://www.puppet.com/docs/puppet/7/lang_comments doesn't say anything about leading spaces

evgeni: added a test for the *one* Pattern I could find using a `#` inside, `Stdlib::Email` :)

lib/kafo/data_type_parser.rb
```diff
@@ -8,9 +8,26 @@ class DataTypeParser
...
+      lines = []
+      last_line = ''
       manifest.each_line do |line|
-        if (type = TYPE_DEFINITION.match(line.force_encoding("UTF-8")))
+        line = line.force_encoding("UTF-8").strip
+        next if line.start_with?('#')
+
+        line = line.split(' #').first.strip
+        if line =~ TYPE_DEFINITION
+          lines << last_line
```
ekohl, line 20: What's the idea behind the `last_line` here? When is that ever relevant for a type definition?

lib/kafo/data_type_parser.rb
```diff
@@ -8,9 +8,26 @@ class DataTypeParser
 
     def initialize(manifest)
       @logger = KafoConfigure.logger
-      @types = {}
+
+      lines = []
+      last_line = ''
```
ekohl, line 13: Using `'' << 'x'` fails with frozen strings so I've learned to avoid that and use `+=` or explicitly mark the string as mutable. Right now the file doesn't have the `frozen_string_literal: true` header, but if it's added, things will break. The easiest is to mark it as mutable:

```suggestion
      last_line = +''
```

ekohl (changes requested):
> This sounds like the correct approach long term, but it also seems like a bigger re-work is required to get that going correctly?

Probably. The patches were written quite some time ago, but never really finished. It did give a massive speed up in building the installer so perhaps it's the excuse I need to properly wrap it up.

But short term this at least allows us to proceed.

lib/kafo/data_type_parser.rb
evgeni: Well, it's a buffer where I store all the lines in one string to drop the newlines.

I renamed the variable now, to make it more obvious

lib/kafo/data_type_parser.rb
evgeni: done

ekohl:
Let's see what https://github.com/theforeman/foreman-installer/pull/898/commits/66f36697bce3e31bb78d9b5071b8366d1c1efc60 does in CI.

ekohl (approved):

PR #369, opened 2023-11-20
Update matrix versions

archanaserver:
Update Ruby ~> 3 and Puppet ~> 8

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -23,11 +23,13 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby:
-          - "2.7"
-        puppet:
-          - "7"
-    
+        include:
+          - ruby: '2.7'
+            puppet: '7'
+          - ruby: '3.0'
+            puppet: '7'
+          - ruby: '3.2'
+            puppet: '8'
     env:
       PUPPET_VERSION: ${{ matrix.puppet }}
       
--- a/lib/kafo/kafo_configure.rb
+++ b/lib/kafo/kafo_configure.rb
@@ -120,7 +120,7 @@ module Kafo
       request_config_reload if applied_total > 0
 
       if ARGV.include?('--migrations-only')
-        verbose = (ARGV.include?('--verbose') || ARGV.include?('-v'))
+        verbose = ARGV.include?('--verbose') || ARGV.include?('-v')
         Logging.setup(verbose: verbose)
         self.class.logger.notice('Log buffers flushed')
         self.class.exit(0)
--- a/lib/kafo/param.rb
+++ b/lib/kafo/param.rb
@@ -161,7 +161,7 @@ module Kafo
 
     def normalize_value(value)
       case value
-        when ::HighLine::String  # don't persist highline extensions
+        when ::HighLine::String # don't persist highline extensions
           value.to_s
         when Array
           value.map { |v| normalize_value(v) }
--- a/lib/kafo/progress_bar.rb
+++ b/lib/kafo/progress_bar.rb
@@ -34,7 +34,7 @@ module Kafo
       @all_lines += 1
 
       # we print every 20th line during installation preparing otherwise only update at EVALTRACE_START
-      update_bar = (@total == :unknown && @all_lines % 20 == 0)
+      update_bar = @total == :unknown && @all_lines % 20 == 0
       force_update = false
 
       if (line_monitor = MONITOR_RESOURCE.match(line))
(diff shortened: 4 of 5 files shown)


ekohl (changes requested):
We still want to keep testing on the old versions, just add the new versions.

.github/workflows/ci.yaml
```diff
@@ -25,8 +25,10 @@ jobs:
       matrix:
         ruby:
           - "2.7"
+          - "3.0"
         puppet:
           - "7"
+          - "8"
```
ekohl, line 31: You can exclude Ruby 2.7 & Puppet 8, since it's an invalid combination:
```suggestion
          - "8"
        exclude:
          - ruby: "2.7"
            puppet: "8"
```

ekohl (commented):
Instead of `[WIP]` you can also mark the PR as a draft (https://github.blog/2019-02-14-introducing-draft-pull-requests/).

.github/workflows/ci.yaml
ekohl: You can also explicitly write out the matrix:
```yaml
include:
  - ruby: '2.7'
    puppet: '7'
  - ruby: '3.2'
    puppet: '8'

**Comment by evgeni** [comment]
The tests execute now and the failure is related.

The internal kafo configure Module is not marked as compatible with Puppet 8:
https://github.com/theforeman/kafo/blob/e8d1a1ea5d21dbd44b35d1ecea29cb296b44cbb4/modules/kafo_configure/metadata.json#L12

Not sure if just bumping this would be sufficient, but certainly worth a try

**Comment by archanaserver** [comment]
@ekohl @evgeni is there a way to check, if the test was still skipped on the previous changes?

**Comment by evgeni** [comment]
> @ekohl @evgeni is there a way to check, if the test was still skipped on the previous changes?

What do you mean by "on the previous changes"? The ones before you pushed today?

You can see previous runs in https://github.com/theforeman/kafo/actions

And they  did not skip the "Tests" part, as there "Rubocop" succeeded.

Seems Rubocop 1.59 is rather new (https://github.com/rubocop/rubocop/releases/tag/v1.59.0) and triggers those errors.
I know they passed with 1.56.4, but no idea if any version inbetween is also good.

We should archive the `Gemfile.lock` generated in theforeman/actions/.github/workflows/rubocop.yml, so we can trace such issues easier.

**Comment by evgeni** [comment]
https://github.com/theforeman/actions/pull/25

**Comment by archanaserver** [comment]
I can't see this rubocop failure in my local, also i just remembered that i've already fix these two cops here https://github.com/theforeman/kafo/blob/master/.rubocop.yml#L156-L160, don't know why it is failing here?

**Comment by ehelms** [comment]
This is what fixed the cops erroring for me:

```
diff --git a/lib/kafo/kafo_configure.rb b/lib/kafo/kafo_configure.rb
index be0a469..9cbdb70 100644
--- a/lib/kafo/kafo_configure.rb
+++ b/lib/kafo/kafo_configure.rb
@@ -120,7 +120,7 @@ module Kafo
       request_config_reload if applied_total > 0
 
       if ARGV.include?('--migrations-only')
-        verbose = (ARGV.include?('--verbose') || ARGV.include?('-v'))
+        verbose = ARGV.include?('--verbose') || ARGV.include?('-v')
         Logging.setup(verbose: verbose)
         self.class.logger.notice('Log buffers flushed')
         self.class.exit(0)
diff --git a/lib/kafo/param.rb b/lib/kafo/param.rb
index 363e16b..cf5047f 100644
--- a/lib/kafo/param.rb
+++ b/lib/kafo/param.rb
@@ -161,7 +161,7 @@ module Kafo
 
     def normalize_value(value)
       case value
-        when ::HighLine::String  # don't persist highline extensions
+        when ::HighLine::String # don't persist highline extensions
           value.to_s
         when Array
           value.map { |v| normalize_value(v) }
diff --git a/lib/kafo/progress_bar.rb b/lib/kafo/progress_bar.rb
index a10810e..ec32f71 100644
--- a/lib/kafo/progress_bar.rb
+++ b/lib/kafo/progress_bar.rb
@@ -34,7 +34,7 @@ module Kafo
       @all_lines += 1
 
       # we print every 20th line during installation preparing otherwise only update at EVALTRACE_START
-      update_bar = (@total == :unknown && @all_lines % 20 == 0)
+      update_bar = @total == :unknown && @all_lines % 20 == 0
       force_update = false
 
       if (line_monitor = MONITOR_RESOURCE.match(line))
```

archanaserver:
@ehelms thanks, i got it!

evgeni (approved):
One could argue this should be 3 commits ("make code compliant with latest rubocop", "mark kafo_configure module as compatible with puppet 8" and "include Ruby 3 and Puppet 8 in test matrix"), but I am not going to nitpick here :)

archanaserver:
Thank you for the review.

PR #372, opened 2024-04-10
Add an option run checks only and exit

ehelms:
The intent is to at least allow two scenarios:

 1. Users that want to be able to run checks before committing to a full install
 2. Allowing foreman-maintain to run the installer checks as part of it's health checks rather than re-implementing the checks

### DIFF
--- a/lib/kafo/kafo_configure.rb
+++ b/lib/kafo/kafo_configure.rb
@@ -193,12 +193,17 @@ module Kafo
         @progress_bar = config.app[:colors] ? ProgressBars::Colored.new : ProgressBars::BlackWhite.new
       end
 
-      unless skip_checks_i_know_better?
-        unless SystemChecker.check
-          puts "Your system does not meet configuration criteria"
+      if checks_only? || !skip_checks_i_know_better?
+        logger = Logger.new('checks')
+        if SystemChecker.check
+          logger.notice("System checks passed")
+        else
+          logger.error("Your system does not meet configuration criteria")
           self.class.exit(:invalid_system)
         end
       end
+
+      self.class.exit(0) if checks_only?
 
       self.class.hooking.execute(:pre_validations)
       if interactive?
@@ -362,6 +367,8 @@ module Kafo
       app_option ['-p', '--profile'], :flag, 'Run puppet in profile mode?',
                  :default => false, :advanced => true
       app_option ['-s', '--skip-checks-i-know-better'], :flag, 'Skip all system checks',
+                 :default => false
+      app_option ['--checks-only'], :flag, 'Run only system checks and exit',
                  :default => false
       app_option ['--skip-puppet-version-check'], :flag, 'Skip check for compatible Puppet versions',
                  :default => false, :advanced => true
--- a/test/acceptance/kafo_configure_test.rb
+++ b/test/acceptance/kafo_configure_test.rb
@@ -66,6 +66,24 @@ module Kafo
       end
     end
 
+    describe '--checks-only' do
+      it 'runs only checks and exits' do
+        FileUtils.mkdir_p "#{INSTALLER_HOME}/checks"
+        FileUtils.cp File.expand_path('../../fixtures/checks/pass/pass.sh', __FILE__), "#{INSTALLER_HOME}/checks"
+        code, out, err = run_command '../bin/kafo-configure --checks-only --verbose --no-colors'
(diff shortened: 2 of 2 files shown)


evgeni:
Didn't we discuss (2) and end up with "this introduces a chicken-egg situation, where people want to run checks *before* they upgrade, but the new checks require the new installer to be present"?

I'm cool with the feature as such, just wonder if we're really getting (2) out of it.

ehelms:
> Didn't we discuss (2) and end up with "this introduces a chicken-egg situation, where people want to run checks _before_ they upgrade, but the new checks require the new installer to be present"?
> 
> I'm cool with the feature as such, just wonder if we're really getting (2) out of it.

Correct - this won't help with new checks in the installer. I was thinking about existing checks that are more persistent? Specifically, I have seen multiple requests to have foreman-maintain health check do a hostname check. I suppose I could just copy that code into foreman-maintain but thought I'd consider this route first and see what others think.

evgeni:
Oh, okay, that it would help with.

ehelms:
@ekohl Any thoughts to adding this feature?

lib/kafo/kafo_configure.rb
```diff
@@ -193,13 +193,18 @@ def execute
         @progress_bar = config.app[:colors] ? ProgressBars::Colored.new : ProgressBars::BlackWhite.new
       end
 
-      unless skip_checks_i_know_better?
-        unless SystemChecker.check
-          puts "Your system does not meet configuration criteria"
+      if checks_only? || !skip_checks_i_know_better?
```
evgeni, line 196: Technically, `checks_only?` is not required here?
Unless you expect people to run `--checks-only --skip-checks-i-know-better`?

ehelms: I don't expect it, but its **possible** and I wanted the outcome to make sense.

evgeni: heh, fair

lib/kafo/kafo_configure.rb
```diff
@@ -193,13 +193,18 @@ def execute
...
+        logger = Logger.new('checks')
+        if !SystemChecker.check
```
ekohl, line 198: Rubocop has a rule to prefer positive tests instead of negative tests. So write
```suggestion
        if SystemChecker.check
```

(and swap the branches)

Would that make more sense?

lib/kafo/kafo_configure.rb
```diff
@@ -193,13 +193,18 @@ def execute
...
+          logger.error("Your system does not meet configuration criteria")
           self.class.exit(:invalid_system)
+        else
+          logger.notice("System checks passed")
```
ekohl, line 202: I'd argue this is `info`, not `notice`. https://en.wikipedia.org/wiki/Syslog#Severity_level says for notice:
> Conditions that are not error conditions, but that may require special handling.

And info:
> Confirmation that the program is working as expected.

ekohl (commented):
Overall I think this makes sense. Some small implementation details inline.

lib/kafo/kafo_configure.rb
ehelms: The problem we always have is that we had to use NOTICE because of how puppet uses INFO and that forced our hand to break some of these rules in order to provide users with useful information. Notice is our default, and I believe that this is a useful output for default state, especially when running checks only or the output feels like nothing actually happened:

```
2024-04-18 14:08:27 [NOTICE] [root] Loading installer configuration. This will take some time.
2024-04-18 14:08:32 [NOTICE] [root] Running installer with log based terminal output at level NOTICE.
2024-04-18 14:08:32 [NOTICE] [root] Use -l to set the terminal output log level to ERROR, WARN, NOTICE, INFO, or DEBUG. See --full-help for definitions.
2024-04-18 14:08:33 [NOTICE] [checks] System checks passed
```

vs.

```
2024-04-18 14:08:27 [NOTICE] [root] Loading installer configuration. This will take some time.
2024-04-18 14:08:32 [NOTICE] [root] Running installer with log based terminal output at level NOTICE.
2024-04-18 14:08:32 [NOTICE] [root] Use -l to set the terminal output log level to ERROR, WARN, NOTICE, INFO, or DEBUG. See --full-help for definitions.
```

ekohl (approved):

evgeni (approved):

PR #375, opened 2024-08-28
Protect parsing of data types from empty lines

lbetz:
If parsing empty rows in data type files the method first returns an nil class so strip doesn't work anymore.

### DIFF
--- a/lib/kafo/data_type_parser.rb
+++ b/lib/kafo/data_type_parser.rb
@@ -13,7 +13,7 @@ module Kafo
       type_line_without_newlines = +''
       manifest.each_line do |line|
         line = line.force_encoding("UTF-8").strip
-        next if line.start_with?('#')
+        next if line.start_with?('#') || line.empty?
 
         line = line.split(' #').first.strip
         if line =~ TYPE_DEFINITION
--- a/test/kafo/data_type_parser_test.rb
+++ b/test/kafo/data_type_parser_test.rb
@@ -48,6 +48,11 @@ module Kafo
       it { _(parser.types).must_equal({"IP"=>"Variant[IPv4,IPv6,]", "IPProto"=>"Variant[TCP,UDP,]"}) }
     end
 
+    describe "parse multiple multiline aliases with empty lines" do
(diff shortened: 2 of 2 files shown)


evgeni:
Thanks Lennart!

The rubocop failure is unrelated (fix in https://github.com/theforeman/kafo/pull/376).

Could you add a test for the empty line behaviour you're fixing?

lib/kafo/data_type_parser.rb
```diff
@@ -15,7 +15,7 @@ def initialize(manifest)
         line = line.force_encoding("UTF-8").strip
         next if line.start_with?('#')
 
-        line = line.split(' #').first.strip
+        line = line.split(' #').first.strip unless line.empty?
```
ekohl, line 18: I had to check this, but what ends up happening is `''.split(' #')` returns `[]` so `[].first` is `nil`.

lib/kafo/data_type_parser.rb
```diff
@@ -15,7 +15,7 @@ def initialize(manifest)
         line = line.force_encoding("UTF-8").strip
         next if line.start_with?('#')
```
ekohl, line 16: I'd prefer to implement this
```suggestion
        next if line.start_with?('#') || line.empty?
```

ekohl (changes requested):

evgeni:
Applied suggestion and added a test.

bastelfreak (approved):

lib/kafo/data_type_parser.rb
```diff
@@ -13,7 +13,7 @@ def initialize(manifest)
       type_line_without_newlines = +''
       manifest.each_line do |line|
         line = line.force_encoding("UTF-8").strip
-        next if line.start_with?('#')
+        next if line.start_with?('#') || line.empty?
```
evgeni, line 16: This won't catch a line full of whitespace (which some editors might leave as is).
```suggestion
        next if line.start_with?('#')
        next if line.strip.empty?
```

WDYT?

evgeni: oh wait, we already strip above, NVM

ekohl (approved):
Btw, long term I'd still prefer to make this depend on puppet-strings to extract the actual data types instead of attempting to maintain yet another parser. Short term this is a good fix.

PR #377, opened 2024-09-24
excape HTML entities (<>&") in HTML and Markdown output

evgeni:
Otherwise params that use those (esp <>) break the rendering of the HTML
output. And because Markdown can contain raw HTML at any place, we
escape it there too.

### DIFF
--- a/test/fixtures/manifests/basic.pp
+++ b/test/fixtures/manifests/basic.pp
@@ -11,6 +11,11 @@
 #                    consisting of 3 lines
 # $typed::           something having it's type explicitly set
 # $multivalue::      list of users
+# $complex_variant:: A Variant type that can be:
+#                    String with:
+#                    '' or 'unmanaged' - Host auth control done elsewhere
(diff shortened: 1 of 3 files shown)


test/fixtures/manifests/basic.pp
```diff
@@ -11,6 +11,11 @@
 #                    consisting of 3 lines
 # $typed::           something having it's type explicitly set
 # $multivalue::      list of users
+# $complex_variant:: A Variant type that can be:
+#                    String with:
+#                    '' or 'unmanaged' - Host auth control done elsewhere
+#                    'ip <List of IPs>' - Allowed IPs/ranges
+#                    Array of strings with ip or host as above
```
evgeni, line 18: We probably can cut this down to a more artificial doc string.
Currently it very much mimics what's in https://github.com/puppetlabs/puppetlabs-apache/blob/main/manifests/mod/status.pp (and breaking our doc builds)

ekohl (approved):

PR #378, opened 2024-10-18
Allow passing extra environment variables to PuppetCommand.format_command

evgeni:


lib/kafo/puppet_command.rb
```diff
@@ -50,11 +50,11 @@ def self.aio_puppet?
       false
     end
 
-    def self.format_command(command)
+    def self.format_command(command, extra_env = {})
       if aio_puppet?
```
ekohl, line 54: Perhaps refactor this method to
```ruby
env = aio_puppet? clean_env_vars : ::ENV.to_h
[env.merge(extra_env), command, :unsetenv_others => false]
```

ekohl:
Note in https://github.com/theforeman/kafo/pull/374 I also started to add tests. Should we build on that?

lib/kafo/puppet_command.rb
evgeni: Considered, but `unsetenv_others` is also different in those branches.

ekohl: Ah yes, now I see why I wrote it that way in the past.

evgeni:
> Note in #374 I also started to add tests. Should we build on that?

Sounds like a plan! You want to (quickly?) finish up #374, or shall I copy over the tests here?

ekohl:
I just rebased it to make RuboCop green.

PR #382, opened 2025-02-11
Use trusted publishers to release the gem

ekohl:
Link: https://community.theforeman.org/t/use-trusted-publishers-to-release-gems/42009


ekohl:
* Created trusted publisher GitHub Actions theforeman/kafo @ .github/workflows/release.yaml (release)
* Cleaned up GH secret
* Removed Rubygems API key

PR #387, opened 2025-10-17
Add Ruby 3.3 / 3.4 support

bastelfreak:
highline 3.0.1 removed the abbrev dependency. This was once a core lib, which was moved to a normal gem in Ruby 3.4. highline 2 still uses it. That means on Ruby 3.4 we need to either add abbrev or update highline.

See https://github.com/search?q=repo%3AJEG2%2Fhighline%20abbrev&type=code

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -16,6 +16,9 @@ gem 'puppet-strings', '>= 1.2.1'
 # Do not pull in Psych 4 since it's incompatible with Puppet
 gem 'rdoc', '< 6.4' if RUBY_VERSION < '3.1'
 
+  # In Ruby 3.4, the gem could moved from default gems to bundled gems and need to be listed explicitly
+gem 'syslog' if RUBY_VERSION >= '3.4'
+
 group :puppet_module do
   gem 'metadata-json-lint'
   gem 'puppetlabs_spec_helper'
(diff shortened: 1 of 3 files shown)


bastelfreak:
Also contains https://github.com/theforeman/kafo/pull/389

bastelfreak:
I can also make a followup PR to update from old Ruby 2.7 to something newer, if you want. Just let me know what your target version is.

bastelfreak:
This also requires https://github.com/theforeman/kafo_wizards/pull/13

evgeni:
@bastelfreak 3.4 is still failing, mind looking into that?

Gemfile
```diff
@@ -16,6 +16,9 @@ gem 'puppet-strings', '>= 1.2.1'
 # Do not pull in Psych 4 since it's incompatible with Puppet
 gem 'rdoc', '< 6.4' if RUBY_VERSION < '3.1'
 
+  # In Ruby 3.4, the gem could moved from default gems to bundled gems and need to be listed explicitly
+gem 'syslog' if RUBY_VERSION > '3.3'
```
ekohl: My mind breaks a bit with doing the mental addition. Won't this work better?
```suggestion
gem 'syslog' if RUBY_VERSION >= '3.4'
```

ekohl (commented):
This remind me, at some point we need to switch over to OpenVox.

PR #395, opened 2026-07-20
Fixes #XXXXX - Serialize concurrent scenario state access with flock

pablomh:
## Summary

When multiple Kafo-based installer commands run concurrently from the same scenario directory (e.g., parallel `foreman-proxy-certs-generate` invocations), three shared resources race:

1. **Scenario config YAML** — `Configuration#initialize` calls `configure_application` which truncates and rewrites the scenario file. A concurrent reader sees an empty or partially-written file, causing `DEFAULT` values (relative paths like `installer_dir: "."`) to be merged and persisted — corrupting the config for all subsequent runs.
2. **Log file rotation** — The `logging` gem's `RollingFile` appender with `truncate: true` renames the log to `._copy_` then to a timestamped backup. Two concurrent instances race on the intermediate `._copy_` file, causing `Errno::ENOENT`.
3. **`last_scenario.yaml` symlink** — `ScenarioManager#link_last_scenario` performs a non-atomic delete followed by a create. Concurrent calls can leave the symlink missing or pointing to a corrupted config.

### Reproduction

```bash
# Run 6+ concurrent foreman-proxy-certs-generate from the same host:
for i in $(seq 1 6); do
  foreman-proxy-certs-generate \
    --certs-tar /tmp/proxy-${i}-certs.tar \
    --foreman-proxy-fqdn proxy-${i}.example.com &
done
wait
```

With 6 concurrent processes, ~2 typically fail. With 10+, most fail. The corruption persists — subsequent sequential runs also fail with "No installation scenario was selected" (rc=26) or "No answer file found" (rc=23) until the scenario config is manually restored.

### Fix

Add `with_scenario_state_lock` using `flock(LOCK_EX)` on a `.kafo-configure.lock` file in the scenario config directory. The lock serializes three critical sections:

- **Initialization:** scenario enable/disable checks, config loading, migrations, and scenario change handling
- **Logging setup:** `RollingFile` appender creation with `truncate: true`
- **Execute save phase:** config persistence, answer file writes, and `last_scenario` symlink updates

Expensive operations (Puppet module parsing, `puppet apply` execution, user hooks) remain outside the lock to allow maximum parallelism. The lock is automatically released when the file descriptor is closed — including on exceptions and `SystemExit`.

## Test plan

- [ ] `bundle exec rake test:unit` passes (no new failures)
- [ ] RuboCop passes on modified files
- [ ] Run concurrent `foreman-proxy-certs-generate` (6+ processes) — no config corruption
- [ ] Verify sequential installer runs are unaffected (no lock contention with a single process)

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -541,6 +541,7 @@ Style/NumericLiteralPrefix:
   Exclude:
     - 'bin/kafofy'
     - 'lib/kafo/configuration.rb'
+    - 'lib/kafo/kafo_configure.rb'
     - 'lib/kafo/logging.rb'
 
 # Offense count: 2
--- a/lib/kafo/kafo_configure.rb
+++ b/lib/kafo/kafo_configure.rb
@@ -9,6 +9,7 @@ end
 
 require 'pty'
 require 'clamp'
+require 'fileutils'
 require 'kafo/color_scheme'
 require 'kafo_parsers/exceptions'
 require 'kafo/exceptions'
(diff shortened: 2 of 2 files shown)


amolpati30:
@pablomh 
I tried this with your patch, but it works sometime or fails for one or two processes with the error below:
```
Execution of '/bin/katello-ssl-tool --gen-client --dir /root/ssl-build --set-hostname proxy-3.example.com --server-cert proxy-3.example.com-puppet-client.crt --server-cert-req proxy-3.example.com-puppet-client.crt.req --server-key proxy-3.example.com-puppet-client.key --no-rpm --password file:/root/ssl-build/katello-default-ca.pwd --set-hostname proxy-3.example.com --set-common-name proxy-3.example.com --ca-cert /root/ssl-build/proxy-3.example.com/katello-default-ca.crt --ca-key /root/ssl-build/proxy-3.example.com/katello-default-ca.key --set-country US --set-state North Carolina --set-city Raleigh --set-org FOREMAN --set-org-unit PUPPET --set-email  --cert-expiration 7300' returned 22: ERROR: web server's SSL certificate generation/signing failed:
```
With further investigation, I noticed that this is a different race condition—not in Kafo. The PR fix is working correctly for the scenario configuration, but the race condition occurs in the SSL certificate generation layer (katello-ssl-tool), which runs during the Puppet apply and intentionally operates outside the Kafo lock.

amolpati30:
When multiple katello-ssl-tool processes run in parallel, they race on these files:
  1. Process A rotates/recreates index.txt while signing a cert
  2. Process B tries to fopen(index.txt, r) at that exact moment → file doesn't exist → failure

amolpati30:
Because it is failing in a different place ([katello-ssl-tool](https://github.com/theforeman/puppet-certs/blob/master/lib/puppet/provider/katello_ssl_tool.rb)), I tried the patch below, and it works.

```
def katello_ssl_tool(*args)
    lock_file_path = File.join(resource[:build_dir], '.katello-ssl-tool.lock')
    File.open(lock_file_path, File::RDWR | File::CREAT, 0600) do |lf|
      lf.flock(File::LOCK_EX)
      Dir.chdir('/root') do
        katello_ssl_tool_command(*args)
      end
    end
end
```
The lock is automatically released when the File.open block ends. This ensures that the certificates are created one at a time, and the file is unlocked after each operation, preventing this failure from occurring.
`ERROR: web server's SSL certificate generation/signing failed:`

pablomh:
I also submitted https://github.com/Katello/katello-certs-tools/pull/51, which solved the initial concurrency problem (I dunno if it's the same issue you're seeing because I'm using the phone :S).

amolpati30:
> I also submitted [Katello/katello-certs-tools#51](https://github.com/Katello/katello-certs-tools/pull/51), which solved the initial concurrency problem (I dunno if it's the same issue you're seeing because I'm using the phone :S).

Yeah, it looks like the same issue applies to serialize_shared_ca in foreman-proxy-certs-generate. I tried both patches, and I don’t see any of the SSL certificate generation/signing errors that I was encountering previously.

amolpati30:
Ack : 🍏
Tested on build : rubygem-kafo-7.8.0-1.20260720185443638018.pr395.1.g303208c.el9.noarch rubygem-kafo-doc-7.8.0-1.20260720185443638018.pr395.1.g303208c.el9.noarch
[51](https://github.com/Katello/katello-certs-tools/pull/51) - With this change.
Result : Success

Scenarios :

1. Run -
  ```
for i in $(seq 1 6); do
    foreman-proxy-certs-generate \
      --certs-tar /tmp/proxy-${i}-certs.tar \
      --foreman-proxy-fqdn proxy-${i}.example.com \
      > /tmp/certs-gen-fix-${i}.log 2>&1 &
  done
  wait
```

2. Check all processes succeeded -
  ```
for i in $(seq 1 6); do
    echo "=== Process $i ==="
    grep -E "Success|Error|exit" /tmp/certs-gen-fix-${i}.log | tail -3
  done
```

3. Verify config is NOT corrupted (installer_dir should NOT be ".") -
```
grep "installer_dir" /etc/foreman-installer/scenarios.d/foreman-proxy-content.yaml
```
4. Verify last_scenario symlink is intact -
```
ls -la /etc/foreman-installer/scenarios.d/last_scenario.yaml
```
5. Verify sequential run still works after concurrent -
  ```
foreman-proxy-certs-generate \
    --certs-tar /tmp/proxy-sequential-certs.tar \
    --foreman-proxy-fqdn proxy-sequential.example.com
```
    
Observation :
- All 6 processes exit 0
- Config YAML is valid — no installer_dir: "." corruption
- last_scenario.yaml symlink exists and points to the correct file
- Sequential run after concurrent works — no leftover lock contention

pablomh:
Ping?

